PR #86, opened 2018-09-19
Bgp

gvarao:
Addressed Harsha's comments in all files
Addressed Wing's comments in all files
Removed old rule file (bgp-neighbor-stats-kpi.rule) which contains all rules as same rules available in separate files
Removed "bgp-neighbor-route-flap-damping.rule" filed due to in correct business requirement


gvarao:
Jenkins please retry a build

gvarao:
Jenkins please retry a build

PR #89, opened 2018-10-17
modified system-proc-ext.yml

AkshayGogeri:
This solves issue with previous system-proc-ext.yml which not used to work in some junos versions


gvarao:
Jenkins please retry a build

PR #108, opened 2018-11-09
OSPF RCA rules

gvarao:
Added descriptions and synopsis to all OSPF and Chip agnostic RCA rules


juniper_official/Protocols/Ospf/ospf-neighbor-state.rule
```diff
@@ -70,4 +70,4 @@ iceberg {
             }
```
ayushgithub: 410
Entering configuration mode
The configuration has been changed but not committed
|/rules/juniper_official/Protocols/Ospf/ospf-neighbor-state.rule:72:(0) syntax error
load complete (1 errors)

gvarao: Addressed and resolved

PR #123, opened 2019-01-14
Adding chassis.networkservices topic

ipmonk:
apologies for the repeated pulls, getting my head around the correct way to do this.

### DIFF
--- a/juniper_official/Chassis/netsvc.py
+++ b/juniper_official/Chassis/netsvc.py
@@ -1,0 +1,769 @@
+""" script to determine whether the network services mode is in sync
+    across config/cli/sysctl/fpc
+"""
+import re
+import sys
+import logging
+import jnpr.junos
+
+# create logger for this script
+LOGGER = logging.getLogger('netsvc.py')
+LOGGER.setLevel(logging.DEBUG)
+CH = logging.StreamHandler()
+CH.setLevel(logging.DEBUG)
+FORMATTER = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                              '%Y-%m-%d %H:%M:%S')
+CH.setFormatter(FORMATTER)
+LOGGER.addHandler(CH)
+
+# create root level logger for debugging pyez/ncclient
+# comment out if not troubleshooting
+# logging.basicConfig(level=logging.INFO)
+
+
+def re_conf(dev, model, routing_engine=None, vc=None, member=None, member_bre=None):
+    """ gets the config for 'show chassis network-services'
+        converts this string to int which maps to the enum in
+        chassis_network_services_t
+        Args:
+            dev(obj): pyez connection to device
+            model(str): router model name
+            routing_engine(str): denotes which RE to run cmd on (opt)
+            vc(bool): True if chassis is part of a VC (opt)
+            member(list): which vc member to connect to (opt)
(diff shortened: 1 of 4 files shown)


ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

juniper_official/Chassis/netsvc.rule
```diff
@@ -0,0 +1,37 @@
...
+        rule netsvc-rule {
+            sensor netsvc-sensor {
+                iAgent {
+                    file netsvc.yml;
+                    table NetSvcTable;
+                    frequency 60s;
+                }
+            }
+            trigger netsvc-sync {
+                term synched {
```
gaddevijay: Default trigger frequency is 10s. Do we want to execute the trigger every 10s when sensor frequency is 60s?

juniper_official/Chassis/netsvc.rule
```diff
@@ -0,0 +1,37 @@
...
+                    then {
+                        status {
+                            color green;
+                        }
+                    }
+                }
+                term else {
+                    when {
+                        matches-with "$netsvc_sync" False;
+                        exists "$proposed_action";
```
gaddevijay: If netsvc_sync is false and there is no proposed action, don't we still flag it as red and say proposed action could not be compiled or something?

juniper_official/Chassis/netsvc.py
```diff
@@ -0,0 +1,769 @@
+""" script to determine whether the network services mode is in sync
```
gaddevijay: So you know anyone, who can review this file from Junos functionality point of view?

juniper_official/Chassis/netsvc.rule
ipmonk: How would a user know what the default is? 

root@20a6061bff84# set iceberg topic chassis.networkservices rule netsvc-rule trigger netsvc-sync frequency ?
Possible completions:
  <frequency>          Frequency or time interval at which the trigger needs to be evaluated. Specify positive integer followed by s/m/h/d/w/y representing seconds/minutes/hours/days/weeks/years. Eg: 2s
[edit]

What is best practice? Should it be the same as sensor frequency? 60s is probably too low for iAgent sensor. I'll discuss with my team.

juniper_official/Chassis/netsvc.rule
ipmonk: proposed action is always a string because we can't reference variables in rules that don’t exist in influx (all referenced fields must be populated) or kapacitor silently does nothing and you see 'no data' in gui.

Currently, if no action is required the string is ''no action required"

juniper_official/Chassis/netsvc.py
ipmonk: Naotaka Fujita is probably the best person, and he's already looked at the script. 
I'm not sure he has a GitHub account to comment here. Will ask.

PR #126, opened 2019-01-14
Committing System/junos-defaults topic on behalf of Sarathi Rao Mungara

ipmonk:


ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

ayushgithub:
jenkins please retry a build

juniper_official/System/junos-defaults.rule
```diff
@@ -0,0 +1,53 @@
...
+                    frequency 30s;
+                }
+            }
+            field default-junos {
+                sensor groupcheck {
+                    path group_name;
+                }
+            }
+        }
+        rule group-check {
```
gaddevijay: Why are two rules required here

juniper_official/System/junos-defaults.rule
```diff
@@ -0,0 +1,53 @@
...
+            rule-frequency 30s;
+            field name-group {
+                reference {
+                    path "/topic[topic-name=system]/rule[rule-name=defaults-group]/field[default-junos='{{group_name}}']/default-junos";
+                    time-range 30s;
```
gaddevijay: If rule(groupcheck) frequency and time-range is same there are chances that we might miss some data at the end of intervals.
My suggestion here is to keep time-range at least 1.5 times the sensor-frequency from where there is being reference

ipmonk: Response from the author:

As I remember, there was a limitation because of which we had to create 2 rules. By using single rule, we were not able to match the “junos-defaults” group. I had sent the challenges faced while creating this playbook with Dan, just sharing for your reference. If they want more details on why we had to use 2 rules, then I can get the specific details from Akshay who had suggested me to use this method. He had shared following bugs with me that time:
 
https://svl-ssd-git.juniper.net/iceberg/jfit-core/issues/110
https://svl-ssd-git.juniper.net/iceberg/jfit-core/issues/149
 
Regarding other point, if they feel rule-frequency and time-range being same, I don’t think there is any issue with keeping it different as well. If they feel, it can cause issues, I am ok with keeping those values different.

gaddevijay: @ipmonk You are right about the two rules. I have resolved that comment already.
This comment above the time-range needs to be changed, else we will end up with no data for the field in some cases.

PR #134, opened 2019-02-06
Added SNMP rules for basic Junos monitoring

remi-blondel:
These rules were initially created to monitor EX3400/4200 series, but should work with every Junos devices.

Here are some important facts:
- Interface-errors rule sometimes shows red squares for every interfaces even though threshold is not reached (quite rare).
- The processes-stats rule retrieve the information correctly, but is showing only one row of the table (The SNMP table for processes is not built the same way as the others, which could explain this).
- I have to put 300s delays for many sensors to avoid DoS of the routing engine.

**I created these rules using the GUI and exported them doing copy/paste from rule builder CLI. The syntax looks good but I had an error when I tested to upload them back. No clues for now**

### DIFF
--- a/community_supplied/cpu-usage.rule
+++ b/community_supplied/cpu-usage.rule
@@ -1,0 +1,87 @@
+iceberg {
+    topic snmp.cpu {
+        rule cpu-usage {
+            keys [ l0index l1index l2index l3index ];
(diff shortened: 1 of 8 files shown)


ayushgithub:
jenkins please retry a build

Issue #122, opened 2019-01-14
Sane default for Software_Version_Value in software-version.rule


ConstrainedRandomVar:
Software version shouldn't be the same as the product. This should be 18.1 or some such..
https://github.com/Juniper/healthbot-rules/blob/master/juniper_official/System/software-version.rule

variable Software_Version_Value {
	value mx960;
	description "Software Version the system should have";
	type string;
}

Regards,

Harsha Lakshmikanth

gvarao:
Harsha..Fixed code is merged, please close the PR.

PR #139, opened 2019-02-22
Modified variable values from mx960/17.X to .*

gvarao:
Addressed issue #122.
Sane default for Software_Version_Value in software-version.rule #122


gvarao:
Re-opening as jenkins issue going to be fixed

PR #175, opened 2019-04-19
Removing VPLS rules

gaddevijay:
Removing VPLS rules as not qualified


gaddevijay:
This has been vetted by Venakta Gajjarapu

PR #183, opened 2019-04-29
adding MAC moves detection rule

mmjn:
Adding MAC moves detection rule


gvarao:
Checks are failing. Please fix and send for review

PR #220, opened 2019-09-09
Healthbot rules for RTSOCK. 

anchal26:
Committing healthbot rules for rtsock.


anchal26:
will raise a new pull request.

PR #236, opened 2019-09-26
Added initial files for two rules

wbellingrath-jnpr:
Trying again after being accepted to relevant alias.

Two initial Rules for Healthbot.
Rule 1 "get-dev-key-state" to check for the status of developer keys.
Rule 2 "test-secure-boot-status" to check if secureboot is running.


wbellingrath-jnpr:
Closing request. Per feedback will place rules under community_supplied and submit a new pull request.

PR #247, opened 2019-10-22
Generic Functions for rule

prashant3167:
Added generic_ functions which consist of udf and uda


juniper_official/System/generic_functions.py
```diff
@@ -0,0 +1,290 @@
...
+previous_time = {} 
+
+# for value function
+prev_bps = {}
+previous_value = {}
+
+
+#UDF
+
+
```
gvarao: Add comments explaining each function

PR #267, opened 2019-12-09
New fixes

gowrisankarvs:
Removed special characters from sys-proc-ext.yml
Increased iagent sensor frequencies and trigger intervals in


gvarao:
jenkins please retry a build

PR #274, opened 2020-01-07
generic_functions.py and system-processes-memory.rule

gowrisankarvs:
1. generic_functions.py updated to handle divide-by-zero issue in decimal_to_percent function.
2. system-processes-memory.rule updated to get the integer values for memory usage.

### DIFF
--- a/juniper_official/System/system-processes-memory.rule
+++ b/juniper_official/System/system-processes-memory.rule
@@ -2,21 +2,21 @@
  * Detects junos processes (daemons) memory utilization threshold breaches and
  * notifies when anomalies are found.
  * Three inputs control detection
- * 
+ *
  *   1) process-name-input, is a regular expression that matches the daemons that you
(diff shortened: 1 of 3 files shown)


juniper_official/System/system-processes-memory.rule
```diff
@@ -25,37 +25,36 @@ healthbot {
...
-                argument res {
-                    mandatory;
-                }
-                argument size {
+            synopsis "System processes memory utilization analyzer";
+            description "Collects system process's memory utilization periodically and notifies anomalies when daemon memory usage exceed threshold";
+            function decimal_to_percent {
+                path generic_functions.py;
+                method decimal_to_percent;
+                argument denominator {
```
gvarao: According to your unit testing results numerator is before denominator. Please do changes accordingly

gowrisankarvs: Modified the rule.

PR #300, opened 2020-04-08
Playbook for syslog rules

d200201123:


juniper_official/Solutions/Pfe-wedge-detection/linecard-pfe-wedge.playbook
```diff
@@ -0,0 +1,7 @@
+healthbot {
```
gvarao, line 1: Please add comments. Refer any default playbooks for same e.g. linecard-kpis

PR #309, opened 2020-05-18
Updated index field

sraju4ug:
index field path was correct to "index" and also made index as non-key


sraju4ug:
committed in wrong branch

PR #323, opened 2020-07-27
Development

gvarao:
- Moved bandwidth monitoring rule to Interfaces folder
- Removed python UDFs and used golang based functions
- Modified two evpns rules as filed asked not to monitor default vlan.


gvarao:
Hi Harsha,

Modified the HB supported version from 3.0.1 to 3.1.0.

Regards,
Venkat(GV)

PR #332, opened 2020-08-07
Resource monitor rules

tkhattar-jnpr:
Rules to fetch and display various FPC and PFE resources.
CLI command "show system resource-monitor summary" and
"show system resource-monitor fpc" provide information about resources
Heap memory, counter memory, CoS utilization, RTT, etc. Added rules to fetch
information about these resources and trigger normal or threshold exceeded.


gvarao:
jenkins please retry a build

PR #338, opened 2020-08-10
Resource monitor rules

tkhattar-jnpr:
Rules to fetch and display various FPC and PFE resources.
CLI command "show system resource-monitor summary" and
"show system resource-monitor fpc" provide information about resources
Heap memory, counter memory, CoS utilization, RTT, etc. Added rules to fetch
information about these resources and trigger normal or threshold exceeded.


tkhattar-jnpr:
Addressed below comments from Gowrisankar:

1) Sensor interval needs to be 60s except some specific case[Some of your rules are configured with 30s]
2) Trigger interval needs to be 1.5*sensor interval [here 90s]. [Most of your rules are configured with 60s]

PR #342, opened 2020-08-11
Patch3

bmadhavi74:
Updated the rule by adding the sensor, fields, variables and trigger conditions.

### DIFF
--- a/community_supplied/Interface/oc-interface-throughput.rule
+++ b/community_supplied/Interface/oc-interface-throughput.rule
@@ -1,3 +1,32 @@
+/*
+* Detects interface throughput utilization and notifies when
+* anomalies are found.
+* Five input control detection
+*
+* 1) in-stats-name, is the interface input traffic counter name
+* whose input octets you would like to monitor. By default it
+* is 'in-octets'.
+*
+* 2) interface-name, is a regular expression that matches the
+* interfaces that you would like to monitor. By default it is '.*',
+* which matches all interfaces. Use something like 'ge.*' to
+* match only gigabit ethernet interfaces.
+*
+* 3) out-stats-name, is the interface output traffic counter name
+* whose output octets you would like to monitor. By default it
+* is 'out-octets'.
+*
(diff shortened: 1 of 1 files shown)


community_supplied/Interface/oc-interface-throughput.rule
```diff
@@ -1,11 +1,350 @@
...
+                    juniper {
+                        operating-system junos {
+                            products EX {
+                                releases 17.3R1 {
+                                    release-support min-supported-release;
+                                    platform all;
+                                }
+                            }
+                            products MX {
+                                releases 15.1R1 {
```
gvarao, line 339: minimum supported version is. 16.2R1.

community_supplied/Interface/oc-interface-throughput.rule
```diff
@@ -1,11 +1,350 @@
...
+            }
+            rule-properties {
+                version 1;
+                contributor juniper;
+                supported-healthbot-version 3.1.0;
+                supported-devices {
+                    juniper {
+                        operating-system junos {
+                            products EX {
+                                releases 17.3R1 {
```
gvarao, line 333: I don't think EX. supports OC

community_supplied/Interface/oc-interface-throughput.rule
```diff
@@ -1,11 +1,350 @@
...
+                            message "$interface-name output bandwidth utilization ($out-bw-util %) is normal. Mbps:$out-mbps Output bytes:$output-bytes";
+                        }
+                    }
+                }
+            }
+            /*
+            * Variables with default values. Default values can be changed
+            * while deploying playbook instance.
+            */
+            variable in-stats-name {
```
gvarao, line 300: No need of templating in, out octets as it is bw rule hence remove in-octets and out-octets variable.

community_supplied/Interface/oc-interface-throughput.rule
bmadhavi74: EX supports a particular platform in OC, so i made the required changes

PR #363, opened 2020-08-28
adding check-ethernet-pause-frame-counters-netconf rule

bmadhavi74:


community_supplied/Interface/check-ethernet-pause-frame-counters-netconf.rule
```diff
@@ -0,0 +1,188 @@
...
+                    file IntStatsTable.yml;
+                    table IntStatsTable;
+                    frequency 60s;
+                }
+            }
+            /*
+            * Fields defined using sensor path. Map the longer sensor names
+            * to the shorter field names used in the rules.
+            */
+            field admin-status {
```
gvarao, line 45: "admin-status" field not being used. You can filter admin down interfaces using where condition in field "interface-name"

gvarao (changes requested):

gvarao (approved):

PR #398, opened 2020-10-01
Update bgp-advertised-routes-kpi.rule

vvikramb:
Changed formula name dynamic-threshold to anomaly detection.


vvikramb:
Changed formula name dynamic-threshold to anomaly detection.

PR #443, opened 2020-10-17
Changed trigger frequency and time range to offset format

bmadhavi74:


gvarao (changes requested):
Please use "offset" instead of just "o"

PR #445, opened 2020-10-19
Added Vendor specific rules

gvarao:


community_supplied/VendorSpecific/Cisco/IOS_XR/check-ios-xr-processes-cpu-netconf.rule
```diff
@@ -0,0 +1,129 @@
...
+    topic system.processes {
+        rule check-ios-xr-processes-cpu-netconf {
+            keys [ pid process-name ];
+            synopsis "System processes cpu analyzer";
+            description "Collects system process's cpu utilization periodically and notifies anomalies when daemon CPU usage exceed threshold";
+            sensor proc {
+                iAgent {
+                    file XRCpuProcessesTable.yml;
+                    table XRCpuProcessesTable;
+                    frequency 60s;
```
ConstrainedRandomVar, line 15: Can you please make this an offset?

community_supplied/VendorSpecific/Paloalto/check-paloalto-processes-cpu-memory-netconf.rule
```diff
@@ -0,0 +1,272 @@
...
+            /*
+             * Sensor configuration to get data from network devices.
+             */
+            sensor systemprocexttable {
+                synopsis "System processes CPU iAgent sensor definition";
+                description "iAgent netconf rpc/cli command sensor to collect data from network device";
+                iAgent {
+                    file PaloaltoSysytemResource.yml;
+                    table PaloaltoSysytemResourceTable;
+                    frequency 60s;
```
ConstrainedRandomVar, line 51: And this..

ConstrainedRandomVar (commented):
Lookd good. Can you please convert to offsets?

community_supplied/VendorSpecific/Cisco/IOS_XR/check-ios-xr-processes-cpu-netconf.rule
gvarao: Offset is only for trigger frequency and time-range and not for sensor in rule. For sensor, we need to configure under ingest settings.

community_supplied/VendorSpecific/Paloalto/check-paloalto-processes-cpu-memory-netconf.rule
gvarao: Offset is only for trigger frequency and time-range and not for sensor in rule. For sensor, we need to configure under ingest settings.

ConstrainedRandomVar (approved):
My bad on the previous comment for moving to frequency offset. Didn't realize that the trigger and time range portions were already done.. Looks great.

PR #490, opened 2021-01-29
Rule to monitor upstr_err from PFE ASIC BT trapstats.

SushilJhawar:
Continous "upstr_err" indicates packets are being dropped by fabric subsyste,m.

This rule runs on all FPCs/linecards.


SushilJhawar:
Will pull latest master & raise new request

PR #516, opened 2021-04-05
Dev: Correcting the facts endpoint

Anam9:
Correcting the facts-collection endpoint. There is a missing "config" keyword after v2 in the API.


Anam9:
@gvarao please check

gvarao (approved):

PR #532, opened 2021-05-07
Patch 100

vvikramb:
Rules added.


gvarao (changes requested):
As we discussed, please address below comments
1.	Try to achieve below in single rule using multiple sensors.
check-queue-red-drop-netconf
check-queue-red-drop-oc
2.	Remove ‘.’ From descriptions
3.	s/ message "queue red dropped packets($queue-counters-total-drop-packets)/ message "Queue red dropped packets($queue-counters-total-drop-packets)/g
4.	s/queue red drop is/ queue red drop count is/g
5.	s/ description "High threshold value." /description "Que red drop count high threshold value"/g
6.	s/ description "Name of the interface."/ description "Input interface-name to monitor"/g

PR #571, opened 2021-09-18
Files

vvikramb:
Predictive Maintenance.

### DIFF
--- a/juniper_official/Predictive-Maintenance/check-optical-temperature-netconf-ml.rule
+++ b/juniper_official/Predictive-Maintenance/check-optical-temperature-netconf-ml.rule
@@ -1,0 +1,211 @@
+/*
+ * Checks for anomalies in optic module temperature 
+ *  
+ */
+healthbot {
+    topic linecard.optical {
+        rule check-optical-temperature-netconf-ml {
+            /*
+             * Monitors anomalies in optic module temperature
+             *
+             * Use interface-name as key for rule.
+             */		
+            keys interface-name;
+            synopsis "Optical module temperature KPI";
+            description "Check if  optical temp is having  anomaly ";
+            /*
+             * Sensor configuration to get data from network devices.
+             */			
+            sensor temperature {
+                iAgent {
+                    file opticalTemperature.yml;
+                    table module_TemperatureTable;
+                    frequency 60s;
+                }
+            }
+            /*
+             * Fields defined with variables and constant values
+             */			
+            field interface-name {
+                sensor temperature {
+                    where "name =~ /{{if-name}}/";
+                    path name;
+                }
+                type string;
+                description "Name of the interface.";
+            }
+            field opt-temp {
+                sensor temperature {
+                    path module-temperature;
+                }
+                type integer;
+                description "Temperature of the Optical Module in degrees Centigrade.";
+            }
+            field opt-temp-anomaly {
+                formula {
+                    anomaly-detection {
+                        algorithm 3sigma;
+                        learning-period 7days;
+                        pattern-periodicity 1days;
+                        field-name "$opt-temp";
+                    }
+                }
+                type integer;
+                description "Anomaly of optic module temperature";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger optical-temperature-anomaly {
+                synopsis "high optical temperature";
+                description "Sets health based on the optical temperature";
+                frequency 1.5offset;
+                /*
+                 * Sets dashboard color to yellow when temperature anomaly is detected and sends 
+                 * notification.
+                 */				
+                term anomaly-present {
+                    when {
+                        equal-to "$opt-temp-anomaly" 1 {
+                            time-range 3offset;
+                        }
+                    }
+                    then {
+                        status {
+                            color yellow;
+                            message "Temperature Anomaly detected for the optical module $interface-name";
+                        }
+                    }
+                }
+                /*
+                 * Sets dashboard color to green when no temperature anomaly is detected and sends 
+                 * notification.
+                 */				
+                term no-anomaly-present {
+                    then {
+                        status {
+                            color green;
+                            message "No Temperature Anomaly detected for the optical module $interface-name  ";
+                        }
+                    }
+                }
+            }
+            /*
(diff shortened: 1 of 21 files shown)


juniper_official/Predictive-Maintenance/check-optical-temperature-netconf-ml.rule
```diff
@@ -0,0 +1,211 @@
...
+                        field-name "$opt-temp";
+                    }
+                }
+                type integer;
+                description "Anomaly of optic module temperature";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger optical-temperature-anomaly {
```
ConstrainedRandomVar, line 59: This is a confusing KPI name.
When the KPI "optical-temperature-anomaly" is green, does it mean that there is an anomaly - as in it is unhealthy; and conversely that it is healthy when red?
Can we rename this to optical-temperature-dt or even just optical-temperature (unless there is a clash there)? 
The more I think about it, the KPI shouldn't even have DT in it.. ML techniques are used to colour the KPI yellow, while breaching static thresholds color them red.

juniper_official/Predictive-Maintenance/check-optics-tx-rx-power-ml.rule
```diff
@@ -0,0 +1,189 @@
...
+                        field-name "$tx-power";
+                    }
+                }
+                type integer;
+                description "tx power anomaly";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger tx-rx-anomaly {
```
ConstrainedRandomVar, line 76: Please pick a more intuitive name for the trigger. Avoid using anomaly in the trigger.
Maybe something like tx-rx-power?

juniper_official/Predictive-Maintenance/check-pfe-normal-discards-ml.rule
```diff
@@ -0,0 +1,136 @@
...
+                sensor pfe-sensor {
+                    path pfe-index;
+                }
+                type string;
+                description "PFE number";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger pfe-bad-route-discard-anomaly {
```
ConstrainedRandomVar, line 74: Please pick a better trigger name.

juniper_official/Predictive-Maintenance/check_adc_temperature-ml.rule
```diff
@@ -0,0 +1,153 @@
...
+ * 
+ *   1) "high-temperature" variable is configured, if needed high values of temperature
+ *       can be checked.
+ *   2) "low-temperature " variable is configured, if needed low values of temperature
+ *       can be checked.
+ *  
+ */
+healthbot {
+    topic adapter.card {
+        rule check_adc_temperature-ml {
```
ConstrainedRandomVar, line 14: Can we avoid using underscores in rule names?

juniper_official/Predictive-Maintenance/check_adc_temperature-ml.rule
```diff
@@ -0,0 +1,153 @@
...
+                sensor adc_card {
+                    path temperature-name;
+                }
+                type string;
+                description "name of the temperature.";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger anomaly {
```
ConstrainedRandomVar, line 85: Please pick a more intuitive KPI name.

juniper_official/Predictive-Maintenance/check_cb_temperature-ml.rule
```diff
@@ -0,0 +1,140 @@
...
+                sensor control_board {
+                    path temperature-name;
+                }
+                type string;
+                description "name of the temperature.";
+            }
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger temperature-anomaly {
```
ConstrainedRandomVar, line 64: Please pick a more intuitive trigger name.

juniper_official/Predictive-Maintenance/check_cb_temperature-ml.rule
```diff
@@ -0,0 +1,140 @@
...
+                iAgent {
+                    file control-board-temperature.yml;
+                    table cb-temp-outlier-Table;
+                    frequency 60s;
+                }
+            }
+            /*
+             * Fields defined with variables and constant values
+             */			
+            field control_board_number {
```
ConstrainedRandomVar, line 28: Please avoid using underscores in field names.

juniper_official/Predictive-Maintenance/check_cb_temperature-ml.rule
```diff
@@ -0,0 +1,140 @@
+/*
+ * Checks for anomalies in temperature for Control Boards 
+ * 
+ */
+healthbot {
+    topic control.board {
+        rule check_cb_temperature-ml {
```
ConstrainedRandomVar, line 7: Please avoid using underscores.

ConstrainedRandomVar (changes requested):
Looks good. Can you please re-send this with the following changes
 - Avoid using underscores in rule names and 
 - Avoid using underscores in field names (unless there is some iAgent limitation, that I vaguely recall - might want to check with Venkat).
 - Where possible don't abbreviate field names - unless we are running into total character length restrictions. For example, opt might be misconstrued for optional rather than optical. Similarly, temp might be read in the mind as temporary or temporal instead of temperature.
 - Keep in mind that the trigger names are the actual KPI names seen in the health view. These should be descriptive, unambiguous, and shouldn't have the word anomaly in them.. For instance, when the KPI "temperatrue-anomaly" is green, does it mean that an anomaly exists? Or conversely, when it is red, does it mean that an anomaly doesn't exist?
 - DT/ML techniques applied to triggers just produce the yellow alarm, while static threshold breaches produce red alarms. So, the KPI shouldn't have ML/DT/anomaly in it wherever possible.
 - From what I recall, learning-period can also take an offset. Might want to check with Vijay on it. Though for your purpose, given that the learning period if sufficiently large compared to the sensor-frequency, this should be okay as is..

Regards,

Harsha Lakshmikanth

PR #581, opened 2021-09-29
Node exporter

vvikramb:
Node Exporter Rules.


juniper_official/Node-Exporter/check-cpu-usage.rule
```diff
@@ -0,0 +1,164 @@
...
+            /*
+             * Anomaly detection logic.
+             */			
+            trigger cpu-usage {
+                frequency 1offset;
+                term above-threshold {
+                    when {
+                        greater-than-or-equal-to "$cpu-used-percentage" "$threshold" {
+                            time-range 3offset;
+                            any;
```
ConstrainedRandomVar, line 129: Don't we want to flag the cpu-usage KPI as red only if the CPU is pegged over a period of time - in which case shouldn't it be all instead of any?

Another option could be to mark this as a yellow trigger (for any) and have a red trigger for all. So long as we don't inundate the operator with too many yellow triggers, even this might be acceptable.

juniper_official/Node-Exporter/check-cpu-usage.rule
```diff
@@ -0,0 +1,164 @@
...
+ * 
+ * Two input control detection.
+ * 
+ *   1) "threshold" is used to check if the used percentage is
+ *      increasing by more than the threshold value.
+ *   2) "freq" is the interval between data points.
+ *  
+ */
+healthbot {
+    topic node.exporter {
```
ConstrainedRandomVar, line 12: Can we come up with a more intuitive topic name? For instance the original ingest name which was node-exporter was changed to server-monitoring. 
Maybe we go with "server monitoring" "server stats" or "host server" or some such. @gvarao - thoughts?

juniper_official/Node-Exporter/check-disk-read-usage.rule
```diff
@@ -0,0 +1,260 @@
...
+            function difference {
+                description "Method that calculates the difference between current and previous readings ";
+                path generic_functions.py;
+                method value_diff;
+                argument key_name {
+                    mandatory;
+                }
+                argument value;
+            }
+            function used-percentage {
```
ConstrainedRandomVar, line 34: Can this be replaced with eval?

juniper_official/Node-Exporter/check-disk-read-usage.rule
```diff
@@ -0,0 +1,260 @@
...
+                    user-defined-function {
+                        function-name used-percentage;
+                        argument total "$frequency";
+                        argument used "$diff-completed";
+                    }
+                }
+                type float;
+                description "Percentage of disk read completed";
+            }
+            field diff-bytes {
```
ConstrainedRandomVar, line 81: Can this be replaced with value-diff?

juniper_official/Node-Exporter/check-disk-read-usage.rule
```diff
@@ -0,0 +1,260 @@
...
+                    user-defined-function {
+                        function-name difference;
+                        argument key_name "$disk-name";
+                        argument value "$completed-total";
+                    }
+                }
+                type integer;
+                description "Difference of previous and current values";
+            }
+            field diff-time-seconds {
```
ConstrainedRandomVar, line 103: Can we use value-diff here as well?

juniper_official/Node-Exporter/check-disk-read-usage.rule
```diff
@@ -0,0 +1,260 @@
...
+                term exceeds-threshold {
+                    when {
+                        greater-than-or-equal-to "$time-seconds-used-percentage" "$threshold" {
+                            time-range 3offset;
+                            any;
+                        }
+                    }
+                    then {
+                        status {
+                            color red;
```
ConstrainedRandomVar, line 225: Should we be flagging disk reads as red at all?
I mean, at best we can indicate that it is anomalous behavior.. But it isn't really errant behavior though.
Should this be yellow?

juniper_official/Node-Exporter/check-network-usage.rule
```diff
@@ -0,0 +1,280 @@
+/*
+ * Checks if network usage fields are exceeding threshold value and raises alarm. 
+ * 
+ * One input control detection.
+ * 
+ *   1) "threshold" is used to check if the used percentage is
+ *      increasinng by more than the threshold value.
```
ConstrainedRandomVar, line 7: typo in increasing.

ConstrainedRandomVar (changes requested):
Made the first pass - but stopped midway since there are some generic comments which need to be applied for all rules.
 - Please leverage value-diff instead of UDF for checking counter values.
 - Please use eval for % computation instead of UDF
 - Please pick a more intuitive topic name.  For instance the original ingest name which was node-exporter was changed to server-monitoring. Maybe we go with "server monitoring" "server stats" or "host server" or some such. @gvarao - thoughts?
 - Finally, be very sensitive to alert fatigue while defining KPIs. Brian Brazil, founder of Robust Perception, has this nice thumbrule.
````
Knowing When Something is Wrong
Alerting is one of the most critical parts of monitoring, but an important question to ask is: What problem is worth waking up an engineer in the middle of the night to look at? It’s tempting to create alerts for anything that’s a red flag or even slightly troublesome, but this can quickly lead to alert fatigue.

Let’s say you’re running a set of user-facing microservices, and you care about the latency of requests. Would central processing unit (CPU) usage on each machine be useful to alert on? The alert will likely flag you that you’re running out of CPU capacity on the machine. It will also have false positives when background processes take a little longer than usual, and false negatives for deadlocks or not having enough threads to use all CPUs.

The CPU is the potential cause of the problem, and high latency is the symptom you are trying to detect. In My Philosophy on Alerting, Rob Ewaschuk points out that there are many potential causes, and it’s difficult to enumerate all of them. It’s better to alert on the symptoms instead, as it results in fewer pages that are more likely to present a real problem worth waking someone up over.
````
In fact, he also has a page on the Prometheus node exporter metrics and says:
"As with most metrics, many of these are only of interest for after-the-fact debugging and are not something you'd put on a dashboard or wake someone up based on."
https://www.robustperception.io/network-interface-metrics-from-the-node-exporter

So, what does it mean for us? 
 - Re-check whether we should be flagging CPU metrics breaching high water mark as red or maybe just as yellow - and that too only if it is prolonged (all v/s any in time-range). 
 - Consider using the disk read/write and other metrics as potential yellows if at all - but not reds.
 - Maybe some of these are interesting metrics and not necessarily KPIs. And to that end, can we create default dashboards in Grafana which would give a view of the behavior over time.
 - Longer term consider how we can observe and track symptoms of malaise in the K8s cluster and use those as KPIs - through synthetic probes - coupled with data from node exporter..

Regards,

Harsha Lakshmikanth

File: juniper_official/Node-Exporter/check-disk-read-usage.rule (hunk 5 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-disk-read-usage.rule]: Changed to value-diff.

File: juniper_official/Node-Exporter/check-network-usage.rule (hunk 7 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-network-usage.rule]: Corrected typo.

File: juniper_official/Node-Exporter/check-disk-read-usage.rule (hunk 6 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-disk-read-usage.rule]: Changed color to yellow.

File: juniper_official/Node-Exporter/check-disk-read-usage.rule (hunk 4 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-disk-read-usage.rule]: Changed to value-diff.

File: juniper_official/Node-Exporter/check-disk-read-usage.rule (hunk 3 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-disk-read-usage.rule]: Replaced with eval.

File: juniper_official/Node-Exporter/check-cpu-usage.rule (hunk 2 above)
Comment (vvikramb) [reply juniper_official/Node-Exporter/check-cpu-usage.rule]: Changed to Server Monitoring.

File: juniper_official/Node-Exporter/check-cpu-usage.rule [hunk 8]
```diff
@@ -0,0 +1,164 @@
...
+                    mandatory;
+                }
+                argument sub_key_name {
+                    mandatory;
+                }
+                argument value {
+                    mandatory;
+                }
+            }
+            function used-percentage {
```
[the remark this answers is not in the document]
vvikramb: Changed to eval.

juniper_official/Node-Exporter/check-cpu-usage.rule
vvikramb: Changed to yellow.

juniper_official/Node-Exporter/check-cpu-usage.rule
```diff
@@ -0,0 +1,164 @@
...
+        rule check-cpu-usage {
+            /*
+             * Monitors cpu usage fields and notifies through alarms if
+             * anomalies are found.
+             *
+             * Use cpu and mode as key for rule.			 
+             */			
+            keys [ cpu mode ];
+            synopsis "CPU Usage KPI";
+            description "Checks CPU total seconds";
```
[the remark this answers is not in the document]
vvikramb: Changed to eval.

ConstrainedRandomVar (approved):
Looks good.

PR #615, opened 2022-01-18
Eval function

vvikramb:
Updated Eval functions.


juniper_official/Kernel/check-too-many-dead-ifstates-iagent.rule
```diff
@@ -44,12 +44,10 @@ healthbot {
...
                 type integer;
             }
-            field ratio {
+            field alive-half-cnt {
                 formula {
-                    user-defined-function {
-                        function-name get-dead-alive-ratio;
-                        argument alive_ifstates_cnt "$alive-cnt";
-                        argument dead_ifstates_cnt "$dead-cnt";
+                    eval {
```
gvarao, line 49: Python function to be removed from rule as well python file. But ensure that no other rule is using python file.
Same applies to all other rules and python files

gvarao (changes requested):

juniper_official/Kernel/check-too-many-dead-ifstates-iagent.rule
vvikramb: Removed the functions from the rules and deleted the pytrhon files that are not being used.

gvarao (approved):

PR #675, opened 2022-11-17
Added new rule to get port data

sarathgollapudi:
The rule parses the port information from the device. The data is use for selecting the Machine Learning model for bad cable inference

### DIFF
--- a/juniper_official/interface/interface-get-port-data.rule
+++ b/juniper_official/interface/interface-get-port-data.rule
@@ -1,0 +1,144 @@
+/*
+ * Get the port information like port name, description and the part number
+ *
+ */
+healthbot {
+    topic interfaces {
+        rule get-port-data {
+            keys port-name;
+            synopsis "Port Data Collector";
+            description "Collects port information from the device";
+            sensor components {
+                synopsis "system components open-config sensor definition";
+                description "/components open-config sensor to collect telemetry data from network device";
+                open-config {
+                    sensor-name /components/;
+                    frequency 60s;
+                }
+            }
+            field description {
+                sensor components {
+                    path /components/component/state/description;
+                }
+                type string;
+            }
+            field part-number {
+                sensor components {
+                    path /components/component/state/part-no;
+                    data-if-missing {
+                        value UNCONNECTED;
+                    }
+                }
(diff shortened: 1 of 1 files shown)


juniper_official/interface/interface-get-port-data.rule
```diff
@@ -0,0 +1,143 @@
...
+            }
+            rule-properties {
+                version 2;
+                contributor juniper;
+                category advanced;
+                supported-healthbot-version 1.0.1;
+                supported-devices {
+                    juniper {
+                        operating-system junos {
+                            products EX {
```
gaddevijay, line 49: jcloud does not support all these models. Please update it with ACX

juniper_official/interface/interface-get-port-data.rule
```diff
@@ -0,0 +1,143 @@
...
+        rule get-port-data {
+            keys port-name;
+            synopsis "Port Data Collector";
+            description "Collects port information from the device";
+            sensor components {
+                synopsis "system components open-config sensor definition";
+                description "/components open-config sensor to collect telemetry data from network device";
+                open-config {
+                    sensor-name /components/;
+                    frequency 60s;
```
gaddevijay, line 15: Do you require this data every 60s? This is mostly static data right?

gaddevijay (changes requested):

juniper_official/interface/interface-get-port-data.rule
sarathgollapudi: Hi Vijay, this is mostly static data, but any change in the hardware need to be updated before the interface data is received in storm topology.
This can be made to 3 minutes.

Let know if this ok.

gaddevijay: Thinking more on this..it does not matter as there are other rules which use /components at 60s.
So do please skip this. No changes required for this comment

juniper_official/interface/interface-get-port-data.rule
gaddevijay: you at PTX also as vArgberg will be used for scale testing

gaddevijay (approved):

PR #677, opened 2022-11-22
Rules modify

vvikramb:
Rules modify

PR #689, opened 2023-01-10
Updated Rules

vvikramb:
Updated Rules


gvarao (approved):

juniper_official/chassis/check-routing-engine-temperature.rule
```diff
@@ -55,7 +55,7 @@ healthbot {
             }
             field routing-engine-temperature {
                 sensor components-oc {
-                    where "/components/component/properties/property/@name == 'RE temperature'";
+                    where "/components/component/properties/property/@name == 'temperature'";
```
gaddevijay, line 58: This change seems to be contradicting https://github.com/Juniper/healthbot-rules/pull/677

PR #694, opened 2023-01-17
Rules Trigger Terms

vvikramb:
Rules Trigger Terms

### DIFF
--- a/juniper_official/chassis/check-chassis-power-fan-temperature.rule
+++ b/juniper_official/chassis/check-chassis-power-fan-temperature.rule
@@ -1,0 +1,150 @@
+/*
+ * Moniters status of chassis temperature, fan and power and notifies when anomalies
+ * are found
+ * Two input controls detection
+ * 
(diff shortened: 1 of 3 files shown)


juniper_official/linecard/check-pfe-discards.rule
```diff
@@ -24,7 +24,7 @@ healthbot {
                 iAgent {
                     file evopfeDiscardStats.yml;
                     table pfeDiscardStats;
-                    frequency 180s;
+                    frequency 60s;
```
gaddevijay, line 27: This should be 180s

juniper_official/chassis/check-chassis-power-fan-temperature.rule
```diff
@@ -1,150 +0,0 @@
-/*
- * Moniters status of chassis temperature, fan and power and notifies when anomalies
```
gaddevijay, line 2: Put this rule back and make sure others remove first

gaddevijay (approved):

PR #698, opened 2023-01-30
RE-Rule-Trigger-Update

vvikramb:
For the rule "hardware.chassis/check-routing-engine-temperature", the trigger conditions for RE temperature and RE CPU temperature are modified to match the below condition.
•	If value is less than high threshold display is green.
•	If value is less than critical threshold display is yellow.
•	If value is greater than or equal to high threshold display is yellow.
•	If value is greater than or equal to critical threshold display is red.

### DIFF
--- a/juniper_official/chassis/check-routing-engine-temperature.rule
+++ b/juniper_official/chassis/check-routing-engine-temperature.rule
@@ -222,12 +222,12 @@ healthbot {
                     }
                 }
                 /*
-                 * Sets color to green when RE CPU temperature is below low threshold ($temp-low-threshold)
+                 * Sets color to green when RE CPU temperature is below low threshold ($low-threshold)
                  * 
                  */				
                 term is-temperature-less-than-low-threshold {
                     when {
-                        less-than "$routing-engine-cpu-temperature" "$temp-low-threshold";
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-routing-engine-temperature.rule
```diff
@@ -124,43 +124,58 @@ healthbot {
...
+                 */  
+                term is-re-temperature-median {
+                    when {
+                        greater-than-or-equal-to "$routing-engine-temperature" "$low-threshold";					
+                        less-than "$routing-engine-temperature" "$high-threshold";
+                    }
+                    then {
+                        status {
+                            color yellow;
+                            message "$routing-engine temperature($routing-engine-temperature degree C) exceeds low threshold";
```
gaddevijay: Are we still calling this as low threshold in the message? I thought we were calling it high or critical based on what is represented in UI graphs

gaddevijay (changes requested):

juniper_official/chassis/check-routing-engine-temperature.rule
vvikramb: Updated trigger messages from low/high threshold to high/critical thresholds.

gaddevijay (approved):

PR #699, opened 2023-01-30
Updated Fan RPM Rule

vvikramb:
For the rule "hardware.chassis/check-fan-state-rpm", the trigger conditions for "rpm-percent" is modified to match the below condition.
• If value is less than high threshold display is green.
• If value is less than critical threshold display is yellow.
• If value is greater than or equal to high threshold display is yellow.
• If value is greater than or equal to critical threshold display is red.

### DIFF
--- a/juniper_official/chassis/check-fan-state-rpm.rule
+++ b/juniper_official/chassis/check-fan-state-rpm.rule
@@ -149,7 +149,7 @@ healthbot {
                  */				
                 term normal {
                     when {
-                        less-than-or-equal-to "$rpm-percent" "$low-threshold";
+                        less-than "$rpm-percent" "$low-threshold";
                     }				
                     then {
                         status {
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -164,13 +164,13 @@ healthbot {
...
-                        greater-than "$rpm-percent" "$low-threshold";
-                        less-than-or-equal-to "$rpm-percent" "$high-threshold";
+                        greater-than-or-equal-to "$rpm-percent" "$low-threshold";
+                        less-than "$rpm-percent" "$high-threshold";
                     }				
                     then {
                         status {
                             color yellow;
-                            message "$fan-name RPM percent $rpm-percent($measurement RPM) is greater than $rpm-low-threshold";
+                            message "$fan-name RPM percent $rpm-percent($measurement RPM) is greater than or equal to $rpm-low-threshold";
```
gaddevijay: There is 'rpm-percent-anomaly' field. We should take care of all the scenarios which include anomaly conditions as well

gaddevijay (changes requested):

juniper_official/chassis/check-fan-state-rpm.rule
vvikramb: Updated Anomaly condition in the terms.

gaddevijay (approved):

PR #703, opened 2023-01-31
Updated check-interface-fec-crc-framing-errors rule

vvikramb:
For the rule "check-interface-fec-crc-framing-errors", the field "input-crc-errors" was changed from type "string" to type "integer".


vvikramb:
The field type for "input-crc-errors" cannot be changed to integer due to backward compatability concerns.Need to add new field as integer type.
The rule is being used by Colombo team for Bad cable data collection.

PR #708, opened 2023-02-01
updating trigger message in check-system-cpu-memory rule

sljuniper131:
for check-psm-power-usage-stage rule, In routing-engine-cpu-utilization and routing-engine-memory-buffer-utilization trigger message has been updated from high to critical.

### DIFF
--- a/juniper_official/system/check-system-cpu-memory.rule
+++ b/juniper_official/system/check-system-cpu-memory.rule
@@ -187,7 +187,7 @@ healthbot {
                      then {
                          status {
                              color red;
-                             message "$routing-engine CPU utilization($re-cpu-utilization) exceeds high threshold($re-cpu-utilization-high-threshold)";
+                             message "$routing-engine CPU utilization($re-cpu-utilization) exceeds critical threshold($re-cpu-utilization-high-threshold)";
                          }
                      }
                  }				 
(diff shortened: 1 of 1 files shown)


juniper_official/system/check-system-cpu-memory.rule
```diff
@@ -187,7 +187,7 @@ healthbot {
                      then {
                          status {
                              color red;
-                             message "$routing-engine CPU utilization($re-cpu-utilization) exceeds high threshold($re-cpu-utilization-high-threshold)";
+                             message "$routing-engine CPU utilization($re-cpu-utilization) exceeds critical threshold($re-cpu-utilization-high-threshold)";
```
gaddevijay, line 190: Below when conditions needs to be changed. 'all' should be changed to any

`                 term is-re-cpu-utilization-middle {
                     when {
                         greater-than-or-equal-to "$re-cpu-utilization" "$re-cpu-utilization-low-threshold" {
                             time-range 3offset;
                             all;
                         }
                         less-than "$re-cpu-utilization" "$re-cpu-utilization-high-threshold" {
                             time-range 3offset;
                         }						 
                     }`

juniper_official/system/check-system-cpu-memory.rule
```diff
@@ -258,7 +258,7 @@ healthbot {
                      then {
                          status {
                              color red;
-                             message "$routing-engine memory buffer utilization($re-memory-buffer) exceed high threshold($re-memory-buffer-high-threshold)";
+                             message "$routing-engine memory buffer utilization($re-memory-buffer) exceed critical threshold($re-memory-buffer-high-threshold)";
```
gaddevijay, line 261: We need to do anomaly change as well.

gaddevijay (changes requested):

gaddevijay:
handled via 715

PR #709, opened 2023-02-01
Draft: Deleted chassis-power-fan-temperature rule

vvikramb:
Deleted "check-chassis-power-fan-temperature" rule and "FanTemperature.yml"


vvikramb:
Post discussion with team, the rule will not be deleted now, but the following action will be taken instead.

check-chassis-power-fan-temperature
      Modification: Delete this rule as this is being tracked by 3 more rules chassis/check-psm-power-usage-state.rule, chassis/check-routing-engine-temperature.rule, chassis/check-fan-state-rpm.rule
Columbo - No Change
Test team - Inform test team about delete of this rule and mentioned why
Khushwant - Remove the trigger name
Vinav - No change
Nikhil - Remove this rule
Venkat - Remove this from the instantiation
Upgrade - There is not way to handle older instantiations
UI - No change
Vikram - Delete the rule   <<< this is in red and should not be done.

PR #710, opened 2023-02-02
updating the trigger messages in check psm temparature rule

sljuniper131:
for check-psm-power-temparature rule, In psm-temperature trigger message has been updated from low to high and high to critical.


juniper_official/chassis/check-psm-temperature.rule
```diff
@@ -97,7 +97,7 @@ healthbot {
                     then {
                         status {
                             color yellow;
-                            message "$psm temperature $temperature is exceeding lower threshold";
+                            message "$psm temperature $temperature is exceeding high threshold";
```
gaddevijay, line 101: Change the when conditions >= low and < high thresholds

gaddevijay (changes requested):

PR #712, opened 2023-02-02
Change the yml file for "check fan state rpm" rule, updated fields

vvikramb:
Deleted "fanEVOTable.yml" and added "fanTable.yml".

Capture Groups formula is used in the rule to collect integer value.

Removed fields :
measurement
rpm-percent

Added fields :
measurements
rpm-measurement
rpm-percents
rpm-percentage


vvikramb:
New MR is opened for the same issue : https://github.com/Juniper/healthbot-rules/pull/714

PR #714, opened 2023-02-03
Fan state rpm feb3

vvikramb:
Added "fanTable.yml".

Modified "check-fan-state-rpm" rule.
Capture Groups formula is used in the rule to collect integer value.

Modified below fields to capture integer value:
measurement
rpm-percent

Added fields :
measurements
rpm-percents


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -24,8 +24,8 @@ healthbot {
              */			
             sensor fan-netconf {
                 iAgent {
-                    file fanEVOTable.yml;
-                    table fanEVOTable;
+                    file fanTable.yml;
```
gaddevijay: Why new file?

juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -35,24 +35,52 @@ healthbot {
...
             field fan-name {
                 sensor fan-netconf {
-                    path Item;
+                    path name;
                 }
                 type string;
                 description "Name of the fan tray and fan number";
             }
-            field measurement {
+            field measurements {
```
gaddevijay, line 43: lets not change the field names as this will impact UI and test

juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -35,24 +35,52 @@ healthbot {
...
+                        field-name "$measurements";
+                        pattern "\d+";
+                    }
                 }
                 type integer;
-                description "RPM measurement";
+                description "RPM Value";				
             }
-            field rpm-percent {
+            field rpm-percents {
```
gaddevijay, line 64: please retain same name

gaddevijay (changes requested):

juniper_official/chassis/check-fan-state-rpm.rule
vvikramb: Updated file name to same as old name (fanEVOTable.yml).

juniper_official/chassis/check-fan-state-rpm.rule
vvikramb: measurement field is still present and uses capture-group formula to derive integer value from measurements.
Measurements is new field which is a string type.

juniper_official/chassis/check-fan-state-rpm.rule
vvikramb: rpm-percent field is still present and uses capture-group formula to derive integer value from rpm-percents.
rpm-percents is new field which is a string type.

juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -150,7 +178,8 @@ healthbot {
                 term normal {
                     when {
                         less-than "$rpm-percent" "$low-threshold";
-                        less-than-or-equal-to "$rpm-percent-anomaly" 0;						
+                        less-than-or-equal-to "$rpm-percent-anomaly" 0;
+                        not-equal-to "$rpm-percent" 0;
```
gaddevijay: Remove these conditions as state will take care of down fan

gaddevijay (approved):

PR #719, opened 2023-02-21
Add missing optic fields to interface error details

skk-JNPR:


juniper_official/interface/interface-error-details.yml
```diff
@@ -28,6 +29,9 @@ intfStatsView:
     fec_uncorr_err: {fec_nccw_count: int}
     fec_corr_err_rate: {fec_ccw_error_rate: int}
```
gaddevijay: This file has too many fields which are not used in the rules. Can you please remove them which are not required by the rules?

gaddevijay (changes requested):

juniper_official/interface/interface-error-details.yml
skk-JNPR: Done

gaddevijay (approved):

PR #734, opened 2023-03-07
updated the when condition in trigger terms

sljuniper131:
updated the when condition in below trigger terms
    1. term is-ospf-hello-receive-not-increasing
    2. term is-ospf-hello-sent-not-increasing

### DIFF
--- a/juniper_official/routing/check-ospf-statistics.rule
+++ b/juniper_official/routing/check-ospf-statistics.rule
@@ -63,6 +63,17 @@
                 synopsis "ospf hello received statistics kpi";
                 description "Sets health based on ospf hello received statistics";
                 frequency 1offset;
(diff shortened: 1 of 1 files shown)


juniper_official/routing/check-ospf-statistics.rule
```diff
@@ -79,14 +77,16 @@
                     }
                 }
                 term is-ospf-hello-receive-not-increasing {
+                    when {
+                        matches-with-previous "$hello-received-5sec";
```
gaddevijay: We should probably do red first and then green for these cases as present logic has some scenarios uncovered
let say threshold is 2 previous value is 10 and present is 11. In this case we will not write anything to trigger table

sljuniper131:
Vijay vikram already made the changes in other PR

PR #735, opened 2023-03-13
updated the frequency value to 1

sljuniper131:


sljuniper131:
Need to raise new MR from updated code

PR #736, opened 2023-03-13
updated the frequency offset value set to 1

sljuniper131:


juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
```diff
@@ -209,7 +209,7 @@ healthbot {
             trigger interface-input-traffic {
                 synopsis "Interface input traffic kpi";
                 description "Sets health based on input traffic exceed threshold";
-                frequency 2offset;
+                frequency 1offset;
```
gaddevijay, line 212: We need remove offset in yellow and red

juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
```diff
@@ -269,7 +269,7 @@ healthbot {
             trigger interface-input-errors {
                 synopsis "Interface input-errors kpi";
                 description "Sets health based on the change in input-errors count";
-                frequency 2offset;
+                frequency 1offset;
```
gaddevijay, line 271: We need remove offset in yellow and red. Same for all triggers

gaddevijay (approved):

PR #765, opened 2023-03-21
updated xslt in yml and ospf sent received rule

vvikramb:
Updated xslt in "OspfStatsTable.yml" and yml table name in "check-ospf-statistics" rule.

### DIFF
--- a/juniper_official/routing/OspfStatsTable.yml
+++ b/juniper_official/routing/OspfStatsTable.yml
@@ -1,0 +1,36 @@
+---
+OspfStatsTable:
+  rpc: get-ospf-statistics-information
+  item: ospf-statistics/packet-statistics[ospf-packet-type="Hello"]
+  key: ospf-packet-type
+  view: OspfStatsView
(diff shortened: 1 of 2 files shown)


juniper_official/routing/OspfStatsTable.yml
```diff
@@ -0,0 +1,36 @@
...
+  rpc: get-ospf-statistics-information
+  item: ospf-statistics/packet-statistics[ospf-packet-type="Hello"]
+  key: ospf-packet-type
+  view: OspfStatsView
+
+OspfStatsView:
+  fields:
+    hello-sent: packets-sent
+    hello-received: packets-received
+    hello-sent-5sec: packets-sent-5seconds
```
gaddevijay: Do we require 5 sec as we are not using them anymore?

same in xslt as well

gaddevijay (changes requested):

gaddevijay (approved):

PR #770, opened 2023-03-27
updated the offset value in check-bgp-neighbor-stats-rule

sljuniper131:


juniper_official/routing/check-bgp-neighbor-stats.rule
```diff
@@ -92,7 +92,7 @@
                     when {
                         increasing-at-least-by-value "$flap-count" {
                             value "$flap-count-threshold";
-                            time-range 3offset;
+                            time-range 2.5offset;
```
gaddevijay, line 97: matches-with-previous should also have 2.5 offset

gaddevijay (changes requested):

juniper_official/routing/check-bgp-neighbor-stats.rule
sljuniper131: Resolved

gaddevijay (approved):

PR #803, opened 2023-04-16
Modified Chassis sensor path

vvikramb:
Modified Chassis Sensor Path for "check-system-power-usage-temp-state" rule.

### DIFF
--- a/juniper_official/chassis/check-system-power-usage-temp-state.rule
+++ b/juniper_official/chassis/check-system-power-usage-temp-state.rule
@@ -291,7 +291,7 @@ healthbot {
                 type int;
             }
             variable slot-no {
-                value 0-1;
+                value 0-9;
                 description "Chassis slot number.";
                 type string;
             }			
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-system-power-usage-temp-state.rule
```diff
@@ -288,6 +290,11 @@ healthbot {
                 description "System remaining power low threshold percentage";
                 type int;
             }
+            variable slot-no {
+                value 0-1;
```
gaddevijay: Why is it restricted to only 0 and 1? Can we broaden the search?

gaddevijay (changes requested):

juniper_official/chassis/check-system-power-usage-temp-state.rule
vvikramb: Updated slot to (0-9).

gaddevijay (approved):

PR #804, opened 2023-05-02
updated FEC in trigger message

vvikramb:
Updated message for interfaces/check-interface-fec-crc-framing-errors rule.

### DIFF
--- a/juniper_official/interface/check-interface-fec-crc-framing-errors.rule
+++ b/juniper_official/interface/check-interface-fec-crc-framing-errors.rule
@@ -113,7 +113,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "fec-corrected-errors $fec-corrected on $interface-name has not increased";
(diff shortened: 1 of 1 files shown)


juniper_official/interface/check-interface-fec-crc-framing-errors.rule
```diff
@@ -151,7 +151,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "fec-uncorrected-errors $fec-uncorrected on $interface-name has not increased";
+                            message "Fec Uncorrected Errors $fec-uncorrected on $interface-name has not increased";
```
gaddevijay: FEC should be capital?

gaddevijay (changes requested):

juniper_official/interface/check-interface-fec-crc-framing-errors.rule
vvikramb: Changed to FEC.

PR #825, opened 2023-06-14
Adding two new rule files for collecting Optical Interface Parameters

harshits158:
Adding below two rule files:
* **1) get-module-temperature.rule**
      To collect below parameters
            1) module-temperature

* **2) get-optical-interface-parameters.rule**
      To collect below parameters:
            1) Interface Name
            2) Lane index
            3) Lane Laser Bias Current
            4) Lane Laser Output Power
            5) Lane Laser Receive Power


juniper_official/interface/get-optical-interface-parameters.rule
```diff
@@ -0,0 +1,94 @@
...
+                sensor sensor_optics {
+                    path "/interfaces/interface/optics/lanediags/lane/@number";
+                }
+                type string;
+                description "Interface lane index";
+            }
+
+            field lane-laser-bias-current {
+                sensor sensor_optics {
+                    path /interfaces/interface/optics/lanediags/lane/lane_laser_bias_current;
```
gaddevijay: Only this path is missing in https://github.com/Juniper/healthbot-rules/blob/jcloud/juniper_official/interface/check-optical-signal-loss-fec-tx-rx-power.rule. We should enhance this rule instead of writing a new one.

juniper_official/interface/get-module-temperature.rule
```diff
@@ -0,0 +1,69 @@
...
+                sensor sensor_optics {
+                    where "/interfaces/interface/@name =~ /{{interface-name}}/";
+                    path "/interfaces/interface/@name";
+                }
+                type string;
+            }
+
+            field module-temperature {
+                sensor sensor_optics {
+                    path /interfaces/interface/optics/module_temp;
```
gaddevijay: Seems like this is already covered in https://github.com/Juniper/healthbot-rules/blob/jcloud/juniper_official/interface/check-optical-temp-thresholds.rule

gaddevijay (changes requested):

sharanyab16:
Jenkins please retry a build

PR #841, opened 2023-11-09
Updated rule properties in "check-isis-adjacency-status" for VMX

vvikramb:
Updated rule properties for "routing.isis/check-isis-adjacency-status" to include "isis-oc-junos" as default ingest for VMX.

### DIFF
--- a/juniper_official/routing/check-isis-adjacency-status.rule
+++ b/juniper_official/routing/check-isis-adjacency-status.rule
@@ -150,6 +150,12 @@
                                     releases 23.2R1 {
                                         release-support min-supported-release;
                                     }
+                                }
+                                platforms vmx {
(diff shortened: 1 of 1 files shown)


juniper_official/routing/check-isis-adjacency-status.rule
```diff
@@ -150,6 +150,12 @@
                                     releases 23.2R1 {
                                         release-support min-supported-release;
                                     }
+                                }
+                                platforms vmx {
```
gaddevijay: This will not work. As the product is not VMX.
Either put default sensors for operating-system junos or add a product for VMX

gaddevijay (changes requested):

gaddevijay (approved):

PR #850, opened 2023-12-08
Updated sensor path to support PTX

vvikramb:
Added sensor path in "hardware.chassis/check-psm-power-usage-state"  to support PTX10008.

### DIFF
--- a/juniper_official/chassis/check-psm-power-usage-state.rule
+++ b/juniper_official/chassis/check-psm-power-usage-state.rule
@@ -373,6 +373,14 @@ healthbot {
                                     }
                                 }
                             }
+                            products PTX {
+                                sensors components-oc-ptx;
+                                platforms PTX10008 {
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-psm-power-usage-state.rule
```diff
@@ -374,6 +398,14 @@ healthbot {
                                     }
                                 }
                             }
+                            products PTX {
```
gaddevijay: Should this be under evo?

gaddevijay (changes requested):

juniper_official/chassis/check-psm-power-usage-state.rule
vvikramb: Updated under evo.

gaddevijay (approved):

PR #853, opened 2023-12-12
Updated no-errors term in ospf-io-stats rule

vvikramb:
Updated no-errors term in "routing.ospf/check-ospf-io-statistics" rule.

### DIFF
--- a/juniper_official/routing/check-ospf-io-statistics.rule
+++ b/juniper_official/routing/check-ospf-io-statistics.rule
@@ -17,6 +17,9 @@
             field ospf-error {
                 sensor ospf-io-statistics-netconf {
                     path ospf-error;
+                    data-if-missing {
+                        value 0;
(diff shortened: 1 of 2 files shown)


juniper_official/routing/check-ospf-io-statistics.rule
```diff
@@ -143,6 +146,13 @@
                                     }
                                 }
                             }
+                            products PTX {
```
gaddevijay, line 157: what about vmx?

gaddevijay (changes requested):

juniper_official/routing/check-ospf-io-statistics.rule
vvikramb: Added for VMX.

gaddevijay (approved):

PR #861, opened 2023-12-19
Removed chassis-state trigger from "hardware.chassis/check-system-power-usage-temp-state"

vvikramb:
Removed "chassis-state" trigger from "hardware.chassis/check-system-power-usage-temp-state".

### DIFF
--- a/juniper_official/chassis/check-system-power-usage-temp-state.rule
+++ b/juniper_official/chassis/check-system-power-usage-temp-state.rule
@@ -1,5 +1,5 @@
 /*
- * Monitors system power  usage,temperature,state and notifies when anomalies are found
+ * Monitors system power usage,temperature and notifies when anomalies are found
  * 
  * Four inputs control detection
  *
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-system-power-usage-temp-state.rule
```diff
@@ -310,6 +264,20 @@ healthbot {
...
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }								
+                                platforms MX304 {
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                            }
```
gaddevijay, line 281: Add vmx product as well

gaddevijay (changes requested):

juniper_official/chassis/check-system-power-usage-temp-state.rule
vvikramb: I have added sensor under Juniper.vmx is not applicable for this rule.

gaddevijay (approved):

PR #871, opened 2024-02-02
Updated fan rule-properties and fan yml

vvikramb:
Updated "hardware.chassis/check-fan-state-rpm" rule and "fanEVOTable" yml file.


vvikramb:
Jenkins please retry a build

PR #873, opened 2024-02-13
PTX rule properties in Fan rule

vvikramb:
Add PTX rule properties in "hardware.chassis/check-fan-state-rpm".


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -331,6 +331,14 @@ healthbot {
                                     }
                                 }								
                             }
+                            products PTX {
+                                sensors fan-netconf;
+                                platforms PTX10008 {
```
gaddevijay, line 336: Do we require for 4k and 16k also?

PR #881, opened 2024-02-13
Added sensor precedence to BGP Advertised routes rule

vvikramb:
"Added sensor precedence to "protocol.bgp/check-bgp-advertised-routes" rule.

### DIFF
--- a/juniper_official/Protocols/Bgp/bgp-advertised-routes-kpi.rule
+++ b/juniper_official/Protocols/Bgp/bgp-advertised-routes-kpi.rule
@@ -224,7 +224,7 @@ healthbot {
                 supported-devices {
                     juniper {
                         operating-system junos {
-						    sensors bgp-sensor;
+                            sensors bgp-sensor;
                             products EX {
                                 platforms EX9200 {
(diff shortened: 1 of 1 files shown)


juniper_official/Protocols/Bgp/bgp-advertised-routes-kpi.rule
```diff
@@ -289,6 +314,16 @@ healthbot {
...
+                                   platforms ACX7100 {
+                                    releases 22.4R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                            }
+                        }						
                     }
                     other-vendor cisco-IOS-XR {
                         vendor-name cisco;
```
gaddevijay, line 330: What is the sensor for Cisco?

gaddevijay (changes requested):

juniper_official/Protocols/Bgp/bgp-advertised-routes-kpi.rule
vvikramb: Updated default sensor under supported-devices.

PR #885, opened 2024-02-14
updated interface optics signal rule properties

vvikramb:
Updated MX and PTX rule properties for "interfaces/check-optical-signal-loss-fec-tx-rx-power" rule.

### DIFF
--- a/juniper_official/interface/check-optical-signal-loss-fec-tx-rx-power.rule
+++ b/juniper_official/interface/check-optical-signal-loss-fec-tx-rx-power.rule
@@ -546,6 +546,13 @@ healthbot {
                                     }
                                 }								
                             }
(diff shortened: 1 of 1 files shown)


juniper_official/interface/check-optical-signal-loss-fec-tx-rx-power.rule
```diff
@@ -546,6 +546,13 @@ healthbot {
                                     }
                                 }								
                             }
+                            products PTX {
```
gaddevijay, line 549: seems like sensor "optical-sensor-oc-junos" is not used in any platform/product...is this correct?

gaddevijay (changes requested):

PR #891, opened 2024-02-14
Dv chassis temp rule

vvikramb:
Updated field routing-engine for "chassis.temperatures/check-chassis-temperature" rule.


juniper_official/Chassis/chassis-temperature.rule
```diff
@@ -145,6 +145,11 @@ healthbot {
                 description "Chassis temperature low threshold: Utilization increase between metrics, before we report anomaly";
                 type int;
             }
+            variable slot-no {
```
gaddevijay, line 148: Why do we want to introduce variable now only for ACX rules.

PR #898, opened 2024-03-22
Added enum to link-state and admins-state fields

vvikramb:
Added enum to "link-state" and "admin-state" fields in "interfaces/check-interface-in-out-errors-traffic-state-flaps" rule.

### DIFF
--- a/juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
+++ b/juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
@@ -67,8 +67,8 @@ healthbot {
                 type string;
                 description "Interface admin status";
                 enumerate __default__ as -1;
-                enumerate down as 0;
-                enumerate up as 1;				
+                enumerate DOWN as 0;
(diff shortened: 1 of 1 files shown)


juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
```diff
@@ -66,6 +66,9 @@ healthbot {
                 }
                 type string;
                 description "Interface admin status";
+                enumerate __default__ as -1;
```
gaddevijay, line 69: Isn't admin and open status have up and down, instead of true and false????

gaddevijay (changes requested):

juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
vvikramb: yes, I have made the changes as Up and down

gaddevijay (approved):

PR #901, opened 2024-03-23
rsvp-te-global-errors-instance-name-change

vvikramb:
Added Ingest to support instance name path from "/network-instances/network-instance/@instance-name" to "/network-instances/network-instance/@name".

### DIFF
--- a/juniper_official/routing/check-te-rsvp-global-errors.rule
+++ b/juniper_official/routing/check-te-rsvp-global-errors.rule
@@ -879,10 +879,10 @@
                         }
                         operating-system junosEvolved {
                             products ACX {
-                                sensors lsp;						
+                                sensors lsp;			
 			                    platforms ACX7024 {
                                     releases 20.1R1 {
-                                        sensors lsp-updated;									
+                                        sensors lsp-updated;
                                         release-support min-supported-release;
                                     }
                                 }
@@ -906,7 +906,7 @@
                                 }
 			                    platforms ACX7509 {
                                     releases 20.1R1 {
-                                        sensors lsp-updated;				
+                                        sensors lsp-updated;
                                         release-support min-supported-release;
                                     }
                                 }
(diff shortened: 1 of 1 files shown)


juniper_official/routing/check-te-rsvp-global-errors.rule
```diff
@@ -515,16 +859,25 @@
                         }
                         operating-system junosEvolved {
                             products ACX {
+                                sensors lsp-updated;
```
gaddevijay: This should be segregated based on the proper ACX models and version

juniper_official/routing/check-te-rsvp-global-errors.rule
```diff
@@ -515,16 +859,25 @@
...
+                                sensors lsp-updated;							
                                 platforms All {
                                     releases 22.3R1 {
                                         release-support min-supported-release;
                                     }
                                 }
                             }
+                            products PTX {
+                                sensors lsp-updated;
+                                platforms PTX10008 {
```
gaddevijay, line 905: Same with PTX

juniper_official/routing/check-te-rsvp-global-errors.rule
```diff
@@ -515,16 +855,27 @@
                         }
                         operating-system junosEvolved {
                             products ACX {
-                                platforms All {
-                                    releases 22.3R1 {
+                                sensors lsp;						
+			                    platforms All {
```
gaddevijay: specific models should be present

gaddevijay (changes requested):

juniper_official/routing/check-te-rsvp-global-errors.rule
```diff
@@ -468,28 +772,87 @@
...
                                     }
                                 }
+                                platforms MX304 {
+                                    releases 20.1R1 {
+                                        sensors lsp-updated;				
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms MX204 {
+                                    releases 20.1R1 {
```
gaddevijay, line 805: Please check the versions. This seems to be an old release. We will have to put all relevant versions if the fix went into multiple versions

gaddevijay (changes requested):

PR #904, opened 2024-04-04
Added enum to fan-state field in hardware.chassis/check-fan-state-rpm rule

vvikramb:
Added enum to field "status" in "hardware.chassis/check-fan-state-rpm" rule.

### DIFF
--- a/juniper_official/chassis/check-fan-state-rpm.rule
+++ b/juniper_official/chassis/check-fan-state-rpm.rule
@@ -97,8 +97,8 @@ healthbot {
                 }
                 type string;
                 description "Status of the fan";
-                enumerate __default__ as -1;
-                enumerate Failed as 0;
+                enumerate __default__ as -99;
+                enumerate Failed as -1;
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -97,6 +97,10 @@ healthbot {
                 }
                 type string;
                 description "Status of the fan";
+                enumerate __default__ as -1;
```
gaddevijay: enumerate __default__ as -99;
                enumerate Failed as -1;
                enumerate Absent as 0;				
                enumerate ok as 1;

gaddevijay (changes requested):

juniper_official/chassis/check-fan-state-rpm.rule
vvikramb: Updated comments.

gaddevijay (approved):

PR #905, opened 2024-04-04
Added enum to psm-state field in hardware.chassis/check-psm-power-usage-state rule

vvikramb:
Added enum to "psm-state" field in "hardware.chassis/check-psm-power-usage-state" rule.

### DIFF
--- a/juniper_official/chassis/check-psm-power-usage-state.rule
+++ b/juniper_official/chassis/check-psm-power-usage-state.rule
@@ -150,6 +150,7 @@ healthbot {
                 }				
                 type string;
                 description "State of power supply module";
+                enumerate __default__ as -1;				
                 enumerate Offline as 0;
                 enumerate Fault as 0;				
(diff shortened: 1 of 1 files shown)


juniper_official/chassis/check-psm-power-usage-state.rule
```diff
@@ -150,6 +150,10 @@ healthbot {
                 }				
                 type string;
                 description "State of power supply module";
+                enumerate __default__ as -1;				
+                enumerate Offline as 0;
+                enumerate Fault as 0;
```
gaddevijay: This should be some other value.. This can be as below. Which means for rest of the rules also we should change to __default__ value to -99

__default__ - -99
Fault - -10
Offline - 0
online - 1

gaddevijay (changes requested):

gaddevijay (approved):

PR #906, opened 2024-04-04
Added enum to state field in hardware.fpc/check-fpc-cpu-memory-state-temp rule

vvikramb:
Added enum to state field in hardware.fpc/check-fpc-cpu-memory-state-temp rule

### DIFF
--- a/juniper_official/linecard/check-fpc-cpu-memory-state-temp.rule
+++ b/juniper_official/linecard/check-fpc-cpu-memory-state-temp.rule
@@ -171,7 +171,7 @@ healthbot {
                 }
                 type string;
                 description "FPC state";
-                enumerate __default__ as -1;				
+                enumerate __default__ as -99;				
                 enumerate Offline as -6;
                 enumerate Offlining as -5;
                 enumerate Onlining as 0;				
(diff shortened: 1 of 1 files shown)


juniper_official/linecard/check-fpc-cpu-memory-state-temp.rule
```diff
@@ -171,6 +171,11 @@ healthbot {
                 }
                 type string;
                 description "FPC state";
+                enumerate __default__ as -1;
```
gaddevijay: enumerate __default__ as -99;				

                enumerate Offline as -6;
                enumerate Offlining as -5;
                enumerate Onlining as 0;				
                enumerate online as 1;

gaddevijay (changes requested):

juniper_official/linecard/check-fpc-cpu-memory-state-temp.rule
vvikramb: Updated Comments.

gaddevijay: Default needs to be update to -99

gaddevijay (changes requested):

gaddevijay (approved):

PR #907, opened 2024-04-04
Added enum to peer-state field in routing.bgp/check-bgp-neighbor-stats rule

vvikramb:
Added enum to "peer-state" field in "routing.bgp/check-bgp-neighbor-stats" rule.

### DIFF
--- a/juniper_official/routing/check-bgp-neighbor-stats.rule
+++ b/juniper_official/routing/check-bgp-neighbor-stats.rule
@@ -62,6 +62,13 @@
                 }
                 type string;
                 description "Sensor field to peer state";
+                enumerate __default__ as -1;				
+                enumerate Idle as 0;
+                enumerate Connect as 0;
+                enumerate Active as 0;
+                enumerate OpenSent as 0;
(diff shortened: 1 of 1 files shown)


juniper_official/routing/check-bgp-neighbor-stats.rule
```diff
@@ -62,6 +62,13 @@
                 }
                 type string;
                 description "Sensor field to peer state";
+                enumerate __default__ as -1;
```
gaddevijay: enumerate __default__ as -99;				
                enumerate Idle as -40;
                enumerate Connect as -30;
                enumerate Active as -20;
                enumerate OpenSent as -10;
                enumerate OpenConfirm as 0;				
                enumerate Established as 1;

gaddevijay (changes requested):

juniper_official/routing/check-bgp-neighbor-stats.rule
vvikramb: Updated Comments.

gaddevijay (approved):

PR #911, opened 2024-04-04
Added enum to "oper-status" field in routing.mpls/check-lsp-state rule

vvikramb:
Added enum to "oper-status" field in "routing.mpls/check-lsp-state" rule.

### DIFF
--- a/juniper_official/routing/check-lsp-state.rule
+++ b/juniper_official/routing/check-lsp-state.rule
@@ -29,7 +29,7 @@
                 }
                 type string;
                 description "LSP operation status";
-                enumerate __default__ as -1;
+                enumerate __default__ as -99;
                 enumerate DOWN as 0;
                 enumerate UP as 1;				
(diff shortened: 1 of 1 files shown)


juniper_official/routing/check-lsp-state.rule
```diff
@@ -29,6 +29,9 @@
                 }
                 type string;
                 description "LSP operation status";
+                enumerate __default__ as -1;
```
gaddevijay: enumerate __default__ as -99;
enumerate DOWN as 0;
                enumerate UP as 1;

gaddevijay (changes requested):

juniper_official/routing/check-lsp-state.rule
vvikramb: Updated Comments.

gaddevijay (approved):

PR #920, opened 2024-05-22
Updated alert type in triggers for fan,psm,re and system rules

vvikramb:
Alert type were added for the below rules :

hardware.chassis/check-fan-state-rpm
hardware.chassis/check-psm-power-usage-state
hardware.chassis/check-routing-engine-temperature
hardware.chassis/check-system-power-usage-temp-state

### DIFF
--- a/juniper_official/chassis/check-fan-state-rpm.rule
+++ b/juniper_official/chassis/check-fan-state-rpm.rule
@@ -156,6 +156,7 @@ healthbot {
              * Anomaly detection logic
              */			
             trigger fan-state {
+                alert-type hardware.state.fan.fan-state;
                 synopsis "Fan State KPI";
                 description "Sets health based on Fan state";			
                 frequency 1offset;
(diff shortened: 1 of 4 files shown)


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -195,6 +196,7 @@ healthbot {
                 }
             }
             trigger rpm-percent {
+                alert-type hardware.rpm.fan.fan-state;
```
gaddevijay: This should be `hardware.rpm.fan.rpm-percent`

juniper_official/chassis/check-psm-power-usage-state.rule
```diff
@@ -183,6 +183,7 @@ healthbot {
              * Anomaly detection logic.
              */			
             trigger psm-power-usage {
+                alert-type hardware.power.psm.usage.psm-power-usage;
```
gaddevijay: this should be `hardware.power.psm.psm-power-usage`

gaddevijay (changes requested):

gaddevijay (approved):

PR #941, opened 2024-06-26
Removed zero-suppression from routing.isis/check-isis-statistics rule fields

vvikramb:
Removed zero-suppression from routing.isis/check-isis-statistics fields and added sensor to support 23.2R1 onwards releases.


sharanyab16:
Jenkins please retry a build

PR #942, opened 2024-06-27
Added ifd and ifl traffic rules 

vvikramb:
Adding interfaces/collect-physical-interface-stats and interfaces/collect-logical-interface-stats rules on behalf of Reinhard.

### DIFF
--- a/juniper_official/interface/collect-logical-interface-stats.rule
+++ b/juniper_official/interface/collect-logical-interface-stats.rule
@@ -12,7 +12,7 @@
  *      interfaces. Use something like '0-10*' to match only 0 to 10 IFLs.
  *
  */
- healthbot {
+healthbot {
     topic interfaces {
         rule collect-logical-interface-stats {
             keys [interface-name sub-interface-index];
--- a/juniper_official/interface/collect-logical-interface-stats.rule
+++ b/juniper_official/interface/collect-logical-interface-stats.rule
@@ -1,5 +1,8 @@
 /*
- * Collect Logical Interface traffic stats.
+ * Collect Logical Interface traffic statistics for pathfinder dynamic topology.
+ *
+ * Req 60: https://paragon-automation.atlassian.net/browse/REQ-60
+ *
  * Two inputs control detection
  *
  *   1) input-ifd-name, is a regular expression that matches the interfaces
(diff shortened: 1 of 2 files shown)


juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -0,0 +1,252 @@
...
+ *
+ *   2) input-ifl-no, is a regular expression that matches the IFL index number that
+ *      you would like to monitor.  By default it's '.*', which matches all
+ *      interfaces. Use something like '0-10*' to match only 0 to 10 IFLs.
+ *
+ */
+ healthbot {
+    topic interfaces {
+		rule collect-logical-interface-stats {
+            keys [ ifl-name interface-name sub-interface-index ];
```
gaddevijay: Making non sensor field(ifl-name) is not a tested feature in EOP. I would suggest to remove this and handle this in business service.

juniper_official/interface/collect-physical-interface-stats.rule
```diff
@@ -0,0 +1,225 @@
...
+            field ifd-ingress-bps {
+                formula {
+                    rate-of-change {
+                        field-name "$ifd-ingress-octets";
+                        multiplication-factor 8;
+                    }
+                }
+                type integer;
+            }
+            field ifd-ingress-octets {
```
gaddevijay: Can we use same naming convention as defined in below rule? At some stage if we want to combine both the rules, it will reduce the changes in business services. In this case it is "in-octets". Same goes with rest of the fields and IFL rule as well

https://github.com/Juniper/healthbot-rules/blob/jcloud/juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule

juniper_official/interface/collect-physical-interface-stats.rule
```diff
@@ -0,0 +1,225 @@
...
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms ACX7348 {
+                                    releases 22.3R1 {
+                                        release-support min-supported-release;
+                                    }
+                                }								
+                            }
+							products PTX {
```
gaddevijay: If any changes are being done to the rule..Please correct all the indentations

gaddevijay (changes requested):

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -0,0 +1,252 @@
...
+ *      that you would like to monitor.  By default it's '.*', which matches
+ *      all interfaces. Use something like 'ge.*' to match only gigabit
+ *      ethernet interfaces. 
+ *
+ *   2) input-ifl-no, is a regular expression that matches the IFL index number that
+ *      you would like to monitor.  By default it's '.*', which matches all
+ *      interfaces. Use something like '0-10*' to match only 0 to 10 IFLs.
+ *
+ */
+ healthbot {
```
gaddevijay, line 18: What is the scale that we are targeting for these rules?

gaddevijay (changes requested):

juniper_official/interface/collect-logical-interface-stats.rule
gaddevijay: Response from Reinhard: For the scale numbers: Again, I haven't found any numbers. For Northstar customers it could be over 100k interfaces. So, I assume the requirement would be in that range.

gaddevijay (approved):

PR #943, opened 2024-06-27
Added rsvp traffic rules 

vvikramb:
Added routing.mpls/collect-rsvp-statson on behalf of Reinhard.

### DIFF
--- a/juniper_official/routing/collect-rsvp-stats.rule
+++ b/juniper_official/routing/collect-rsvp-stats.rule
@@ -139,16 +139,16 @@
                                         release-support min-supported-release;
                                     }
                                 }
+                                platforms PTX10008 {
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }								
                                 platforms PTX10016 {
                                     releases 22.2R3 {
                                         release-support min-supported-release;
                                     }
                                 }								
-                                platforms PTX10008 {
-                                    releases 22.2R3 {
-                                        release-support min-supported-release;
-                                    }
-                                }
                             }
                         }
                     }
--- a/juniper_official/routing/collect-rsvp-stats.rule
+++ b/juniper_official/routing/collect-rsvp-stats.rule
@@ -14,7 +14,7 @@
  healthbot {
     topic routing.mpls {
         rule collect-rsvp-stats {
-            keys lsp-name;
+            keys [ lsp-name lsp-stats-counter ];
             description "Collects LSP stats periodically";
             sensor lsp {
                 open-config {
@@ -126,26 +126,24 @@
                                     }
                                 }
                             }
-                            products PTX {
-                                platforms PTX1000 {
-                                    releases 17.4R1 {
-                                        release-support min-supported-release;
-                                    }
-                                }
-                                platforms PTX10000 {
-                                    releases 19.411 {
-                                        release-support min-supported-release;
-                                    }
-                                }
-                                platforms PTX5000 {
-                                    releases 17.4R1 {
-                                        release-support min-supported-release;
-                                    }
-                                }
-                            }
                         }
                         operating-system junosEvolved {
 							products PTX {
+                                platforms PTX10001 {
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms PTX10004 {
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms PTX10016 {
+                                    releases 22.2R3 {
+                                        release-support min-supported-release;
+                                    }
+                                }								
                                 platforms PTX10008 {
                                     releases 22.2R3 {
                                         release-support min-supported-release;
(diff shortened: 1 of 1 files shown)


juniper_official/routing/collect-rsvp-stats.rule
```diff
@@ -0,0 +1,200 @@
...
+            field lsp-stats-bytes {
+                sensor lsp {
+                    path /mpls/lsps/constrained-path/tunnels/tunnel/state/counters/bytes;
+                }
+                type integer;
+                description "LSP bytes counter";
+            }
+            field lsp-stats-counter {
+                sensor lsp {
+                    path /mpls/lsps/constrained-path/tunnels/tunnel/state/counters/name;
```
gaddevijay, line 51: What is the significance of this? Would there be different counter names? Any documentation? or what values can come here?

gaddevijay: Could not find these path's in telemetry explorer

juniper_official/routing/collect-rsvp-stats.rule
```diff
@@ -0,0 +1,200 @@
...
+ *
+ */
+ healthbot {
+    topic routing.mpls {
+        rule collect-rsvp-stats {
+            keys lsp-name;
+            description "Collects LSP stats periodically";
+            sensor lsp {
+                open-config {
+                    sensor-name /junos/services/label-switched-path/usage/;
```
gaddevijay, line 21: In already existing rsvp rule "
https://github.com/Juniper/healthbot-rules/blob/jcloud/juniper_official/routing/check-rsvp-neighbor-state.rule
" a different sensor path is used. Is there a way that we can use the sensor path for both the rules? [6:40 pm] Vijay Gadde
Or in these cases the these sensors send completely different data?

gaddevijay (changes requested):

juniper_official/routing/collect-rsvp-stats.rule
```diff
@@ -0,0 +1,200 @@
...
+ *
+ *   1) input-lsp-name, is a regular expression that matches the lsp name
+ *      that you would like to monitor.  By default it's '.*', which matches
+ *      all lsps. Use something like 'Silver*' to match only lsp names starting
+ *      with 'Silver'. 
+ *
+ */
+ healthbot {
+    topic routing.mpls {
+        rule collect-rsvp-stats {
```
gaddevijay, line 16: What is the scale that we targeting for this rule?

gaddevijay (changes requested):

juniper_official/routing/collect-rsvp-stats.rule
gaddevijay: Response from Reinhard: I haven't found any number of required LSPs to be supported. In Northstar, we have customers who have 100k + lsps to be managed, but the bulk has less than 100k.
  Having said that, I would go with 100k for now. This is not per device, but for the entire network. I believe this number will only be relevant if existing Northstar customers upgrade to EOP.

juniper_official/routing/collect-rsvp-stats.rule
gaddevijay: As per mail from Reinhard lisp-name also should be in the key

juniper_official/routing/collect-rsvp-stats.rule
```diff
@@ -0,0 +1,173 @@
...
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms PTX5000 {
+                                    releases 17.4R1 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                            }
+                            products QFX {
```
gaddevijay: Please remove QFX also as Reinhard only confirmed MX and PTX

juniper_official/routing/collect-rsvp-stats.rule
```diff
@@ -0,0 +1,173 @@
...
+                                        release-support min-supported-release;
+                                    }
+                                }
+                                platforms MX960 {
+                                    releases 19.2R1 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                            }
+                            products PTX {
```
gaddevijay, line 131: Should we add PTX in EVO?

gaddevijay (changes requested):

juniper_official/routing/collect-rsvp-stats.rule
vvikramb: yes, lsp-name is key.

juniper_official/routing/collect-rsvp-stats.rule
vvikramb: ok, removed QFX.

juniper_official/routing/collect-rsvp-stats.rule
gaddevijay: My comment was not clear "lsp-stats-counter " should also be key

juniper_official/routing/collect-rsvp-stats.rule
vvikramb: Added PTX10001,PTX10004,PTX10008,PTX10016 for EVO.

juniper_official/routing/collect-rsvp-stats.rule
vvikramb: Added "lsp-stats-counter" also as the key.

gaddevijay (approved):

PR #945, opened 2024-07-16
updated rules with enum

nchiru:
added enums in below rules
lldp-session-state.rule
ethernet-oam-lfm.rule
ethernet-oam-lfm.yml
bfd-session-state.rule
evpn-instance-check.rule

### DIFF
--- a/juniper_official/lldp/lldp-session-state.rule
+++ b/juniper_official/lldp/lldp-session-state.rule
@@ -67,7 +67,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "LLDP interface $interface-name status is Up";
+                            message "LLDP status on interface $interface-name is UP";
                         }
                     }
                 }
@@ -78,7 +78,7 @@ healthbot {
                     then {
                         status {
                             color red;
-                            message "Interface-$interface-name lldp session down";
+                            message "Interface $interface-name lldp session down";
                         }
                     }
                 }
--- a/juniper_official/lldp/lldp-session-state.rule
+++ b/juniper_official/lldp/lldp-session-state.rule
@@ -67,7 +67,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "LLDP status on interface $interface-name is UP";
+                            message "LLDP status on interface $interface-name is Enabled";
                         }
                     }
                 }
@@ -78,7 +78,7 @@ healthbot {
                     then {
                         status {
                             color red;
-                            message "Interface $interface-name lldp session down";
+                            message "LLDP status on interface $interface-name is Disabled";
                         }
                     }
                 }
--- a/juniper_official/lldp/lldp-session-state.rule
+++ b/juniper_official/lldp/lldp-session-state.rule
@@ -39,6 +39,8 @@ healthbot {
                 }
                 type string;
                 description "Local interface LLDP state";
+                enumerate false as 0;
+                enumerate true as 1;
             }
             field interface-name {
                 sensor lldp-sensor {
@@ -54,6 +56,22 @@ healthbot {
                 }
                 type string;
                 description "LLDP neighbor state";
+                enumerate false as 0;
+                enumerate true as 1;
+            }
+            field neighbor-system-name {
+                sensor lldp-sensor {
+                    path /lldp/interfaces/interface/neighbors/neighbor/state/system-name;
+                }
+                type string;
+                description "LLDP neighbor system name";
+            }
+            field neighbor-mac-id {
+                sensor lldp-sensor {
+                    path /lldp/interfaces/interface/neighbors/neighbor/state/chassis-id;
+                }
+                type string;
+                description "LLDP neighbor chassis id";
             }
             /*
              * Anomaly detection logic.
@@ -73,7 +91,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "Interface $interface-name lldp session is up";
+                            message "LLDP interface $interface-name status is Up, neighbor details: system:$neighbor-system-name, mac:$neighbor-mac-id";
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


juniper_official/lldp/lldp-session-state.rule
```diff
@@ -67,13 +62,12 @@ healthbot {
...
                     when {
-                        matches-with "$neighbor-state" true;
-                        matches-with "$if-lldp-state" true;
+                        matches-with "$interface-lldp-state" true;
                     }
                     then {
                         status {
                             color green;
-                            message "Interface $interface-name lldp session is up";
+                            message "LLDP interface $interface-name status is Up";
```
gaddevijay: Change this to "LLDP status on interface $interface-name is UP".
LLDP status is coming as true and we are displaying message as UP. Is this correct? Should we say that LLDP on interface is enabled or something like that?

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,6 +27,26 @@ healthbot {
                 }
                 type string;
                 description "Discovery state for a given interface";
+                enumerate Passive Wait as 0;
+                enumerate Send Any as 1;
+                enumerate Send Local Remote as 1;
```
gaddevijay: We cannot have 2 values representing 1.

Also Send Local Remote and Send Local Remote Ok are good states or bad states? if they are bad states please keep it less than 1

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,6 +27,26 @@ healthbot {
...
+                enumerate Send Local Remote Ok as 3;
+            }
+            field lfmd-status {
```
gaddevijay: New field not used in triggers are not stored in TSDB. Why are these being added?

juniper_official/routing/bfd-session-state.rule
```diff
@@ -35,13 +35,19 @@ healthbot {
                 }
                 type string;
                 description "BFD session state";
+                enumerate Failing as -1;
```
gaddevijay: to states can have same value -1

juniper_official/vpn/evpn/evpn-instance-check.rule
```diff
@@ -30,6 +30,8 @@
                 }
                 type string;
                 description "Mode of evpn Interface";
+                enumerate single-homed as 0;
```
gaddevijay: Is single homed a bad state?
If it is a good state it should be >=1

juniper_official/lldp/lldp-session-state.rule
gaddevijay: Also line 81 `message "Interface-$interface-name lldp session down";`
Please put a space after "Interface" and remove "-"

gaddevijay (changes requested):

juniper_official/lldp/lldp-session-state.rule
nchiru: updated as per the comments

juniper_official/oam/ethernet-oam-lfm.rule
nchiru: these are good states. updated with positive values.

juniper_official/routing/bfd-session-state.rule
nchiru: updated with negative values

juniper_official/vpn/evpn/evpn-instance-check.rule
nchiru: single homed is the default value and is good state. updated with positive value

gvarao (approved):

juniper_official/lldp/lldp-session-state.rule
gaddevijay: You did not answer this questions "LLDP status is coming as true and we are displaying message as UP. Is this correct? Should we say that LLDP on interface is enabled or something like that?"

Please check with Venkat and comment if this is ok or not

gaddevijay: Line 81 message format also needs to be corrected as you have corrected message in line 70

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,20 +27,34 @@ healthbot {
                 }
                 type string;
                 description "Discovery state for a given interface";
+                enumerate Passive Wait as 0;
+                enumerate Send Any as 1;
```
gaddevijay: Are you sure that enum names will work without double quotes?

I tried to load into terminal and getting below error.

         terminal:24:(38) syntax error: Wait
    'enumerate Passive Wait as 0;'
      syntax error

gaddevijay (changes requested):

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,20 +27,34 @@ healthbot {
                 }
                 type string;
                 description "Discovery state for a given interface";
+                enumerate "Passive Wait" as 0;
+                enumerate "Send Any" as 1;
```
gaddevijay, line 31: I just spoke to Venkat and as per him 2 and 3 are not good states. So it has to be less than 1

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,20 +27,34 @@ healthbot {
...
+                enumerate "Send Local Remote" as 2;
+                enumerate "Send Local Remote Ok" as 3;
             }
-            field lfm-interface-name {
+            field lfmd-status {
```
gaddevijay: change `lfmd-status` to `lfm-status`

gaddevijay (changes requested):

juniper_official/oam/ethernet-oam-lfm.rule
gvarao: Sorry vijay, i misled. i just gone though document 2 and 3 are positive cases. So we can keep as is now.

gaddevijay (approved):

PR #957, opened 2024-07-31
updated lfm discovery state enums and trigger msg

nchiru:
updated enums as per the below link https://opengrok.juniper.net/source/xref/DEV_COMMON_BRANCH/junos/usr.sbin/lfmd/lfmd_discovery.c?r=1178985#291

### DIFF
--- a/juniper_official/oam/ethernet-oam-lfm.rule
+++ b/juniper_official/oam/ethernet-oam-lfm.rule
@@ -27,28 +27,28 @@ healthbot {
                 }
                 type string;
                 description "Discovery state for a given interface";
-                enumerate "Fault" as -1
-                enumerate "Active Send Local" as 3
-                enumerate "Passive Wait" as 4
-                enumerate "Send Local Remote" as 5
-                enumerate "Send Local Remote OK" as 6
-                enumerate "Send Any" 7
-                enumerate "Lost Link Timer Expired" as -2
-                enumerate "Link Down" as 0
-                enumerate "Link Up Active" as 1
-                enumerate "Link Up Passive" as 2
-                enumerate "Info PDU Received" as 8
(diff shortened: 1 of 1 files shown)


juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -27,10 +27,28 @@ healthbot {
                 }
                 type string;
                 description "Discovery state for a given interface";
-                enumerate "Passive Wait" as 0;
-                enumerate "Send Any" as 1;
-                enumerate "Send Local Remote" as 2;
-                enumerate "Send Local Remote Ok" as 3;
+                enumerate "Fault" as -1;
+                enumerate "Active Send Local" as 3;
+                enumerate "Passive Wait" as 4;
```
gaddevijay, line 32: Do not change enums for already present states

gaddevijay (changes requested):

juniper_official/oam/ethernet-oam-lfm.rule
```diff
@@ -67,15 +85,15 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "$lfm-interface-name LFM discovery-state: $lfm-discovery-state; lfm-status: $lfm-status; lfm-peer-address: $lfm-peer-address";
+                            message "$lfm-interface-name LFM discovery-state: $lfm-discovery-state; lfm-status: $lfm-status";
```
gaddevijay, line 88: If peer address is not used anywhere can we remove the field? Please confirm with Venakt.

gaddevijay (changes requested):

juniper_official/oam/ethernet-oam-lfm.rule
gaddevijay: Also can you confirm that state anything > 1 is a good state?

PR #967, opened 2024-08-13
Updated yml file and "hardware.system/check-ntp-synchronization-status" rule

vvikramb:
Updated yml file and "hardware.system/check-ntp-synchronization-status" rule.


vvikramb:
As per GNATS 1829823, AT10 

* This change is applicable starting releases 24.3 where we implemented RLI 57442.

* In addition to ODL annotation implementation, we had to refine and re-organise certain ODL tags as per UI guidelines.

* This RLI 57442 w.r.t ODL annotation is only applicable to EVO.

* But, this additional change in the RLI context w.r.t refining and re-organising of ODL tags is applicable to both JUNOS and EVO.

* The backend XML tags were refined and not the original CLI rendering data for "show ntp status" and "show ntp associations" and thus this is not categorised as a CVBC.

gaddevijay (approved):

PR #973, opened 2024-08-30
Updated "lldp/check-lldp-session" rule

vvikramb:
Updated lldp/check-lldp-session rule


juniper_official/lldp/lldp-session-state.rule
```diff
@@ -23,62 +23,69 @@ healthbot {
...
             }
             /*
              * Fields defined using sensor path. Map the longer sensor names
              * to the shorter field names used in the rules.
              */
-            field lldp-enabled {
+            field remote-system-name {
                 sensor lldp-sensor {
-                    path /lldp/interfaces/interface/state/enabled;
+                    path remote-system-name;
```
gaddevijay: Can one interface have multiple remote-system?

juniper_official/lldp/lldp-session-state.rule
```diff
@@ -23,62 +23,69 @@ healthbot {
...
+                term is-lldp-neighbor-discovered {
                     when {
-                        matches-with "$lldp-enabled" true;
+                        exists "$remote-system-name";
                     }
                     then {
                         status {
                             color green;
-                            message "LLDP status on interface $interface-name is Enabled";
+                            message "LLDP neighbor state of $interface-name is UP, remote-system-name:$remote-system-name lldp-remote-port-id:$lldp-remote-port-id";
```
gaddevijay: Can LLDP neighbours have different state? If yes, then we cannot use UP. 
Can you a LLDP cli output?

gaddevijay: message needs to changed so that we don't give an impressions to user that state of LLDP neighbour is UP or DOWN

gaddevijay (changes requested):

juniper_official/lldp/lldp-session-state.rule
gvarao: Thanks Vijay for review. No, there will be one remote-system per interface

juniper_official/lldp/lldp-session-state.rule
gvarao: There are no states for LLDP neighbors, if neigbhor exists then discovered else it will be blank. hence created own states.

gaddevijay: LLDP neighbor detected on $interface-name with remote system name $remote-system-name and remote port $lldp-remote-port-id

juniper_official/lldp/lldp-session-state.rule
```diff
@@ -23,62 +23,69 @@ healthbot {
...
                 /*
                  * Defaults color to red and notify anomaly.
                  */
-                term lldp-session-down {
+                term lldp-neighbor-undiscovered {
                     then {
                         status {
                             color red;
-                            message "LLDP status on interface $interface-name is Disabled";
+                            message "LLDP neighbor state of $interface-name is DOWN";
```
gaddevijay: if system id and remote port is not generated...then how will we come here?

gaddevijay: Since there are no fields. this point is invalid and will be dropped

juniper_official/lldp/lldp-session-state.rule
```diff
@@ -23,62 +23,61 @@ healthbot {
...
                 description "Sets health based on lldp session state changes and notify anomaly.";
                 frequency 1offset;
                 /*
                  * Sets color to green when LLDP neighbor state is true.
                  */
-                term is-lldp-session-up {
+                term is-lldp-neighbor-discovered {
                     when {
-                        matches-with "$lldp-enabled" true;
+                        greater-than-or-equal-to "$lldp-neighbor-status" 1;
```
gaddevijay: lldp-neighbor-status is not defined

juniper_official/lldp/lldp-session-state.rule
```diff
@@ -23,62 +23,61 @@ healthbot {
...
                 /*
                  * Defaults color to red and notify anomaly.
                  */
-                term lldp-session-down {
+                term lldp-neighbor-undiscovered {
                     then {
                         status {
                             color red;
-                            message "LLDP status on interface $interface-name is Disabled";
+                            message "$lldp-neighbor-count lldp neighbors discovered on $interface-name";
```
gaddevijay, line 80: In rule properties, please add all the models that are supported by EOP. Or at least put default Junos and EVO versions. Platform ALL is not valid

gaddevijay (changes requested):

gaddevijay (approved):

PR #993, opened 2024-11-17
Updated ex4300 in rule properties for ISIS Rules

vvikramb:
routing.isis/check-isis-adjacency-status
routing.isis/check-isis-flap-detection
routing.isis/check-isis-statistics

### DIFF
--- a/juniper_official/routing/check-isis-statistics.rule
+++ b/juniper_official/routing/check-isis-statistics.rule
@@ -491,6 +491,13 @@
                                     }
                                 }
                             }
+                            products EX {
+                                platforms EX4300-48MP {
+                                    releases 22.4R2.8 {
+                                        release-support min-supported-release;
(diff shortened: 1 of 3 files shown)


juniper_official/routing/check-isis-statistics.rule
```diff
@@ -491,6 +491,13 @@
                                     }
                                 }
                             }
+                            products EX {
```
gaddevijay, line 494: Can we put sensors also at EX level? Same comment for all 3 rules

gaddevijay (changes requested):

juniper_official/routing/check-isis-statistics.rule
vvikramb: Updated sensor under EX product for all 3 rules.

PR #994, opened 2024-11-17
Updated ex4300 in rule properties for BGP Rules

vvikramb:
routing.bgp/check-bgp-neighbor-stats
routing.bgp/check-bgp-neighbor-prefixes

### DIFF
--- a/juniper_official/routing/check-bgp-neighbor-stats.rule
+++ b/juniper_official/routing/check-bgp-neighbor-stats.rule
@@ -168,8 +168,8 @@
                     juniper {
                         operating-system junos {
                             products EX {
-                                platforms All {
-                                    releases 15.1R1 {
+                                platforms EX4300-48MP {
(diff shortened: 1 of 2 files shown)


juniper_official/routing/check-bgp-neighbor-stats.rule
```diff
@@ -168,8 +168,8 @@
                     juniper {
                         operating-system junos {
                             products EX {
-                                platforms All {
```
gaddevijay, line 171: Add sensor at product level

gaddevijay (changes requested):

juniper_official/routing/check-bgp-neighbor-stats.rule
vvikramb: Added sensor at EX Product level for both the rules.

PR #1001, opened 2024-11-19
Enhanced BGP rule with instance-name

gvarao:
Enhanced BGP rule with instance-name

### DIFF
--- a/juniper_official/routing/check-bgp-neighbor-prefixes.rule
+++ b/juniper_official/routing/check-bgp-neighbor-prefixes.rule
@@ -201,37 +201,29 @@
                     sensors bgp-oc;
                     juniper {
                         operating-system junos {
-                            products EX {
-                                platforms All {
-                                    releases 15.1R1 {
-                                        release-support min-supported-release;
-                                    }
-                                }
-                            }
                             products MX {
-                                platforms All {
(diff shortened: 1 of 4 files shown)


juniper_official/routing/check-bgp-neighbor-prefixes.rule
```diff
@@ -23,6 +23,18 @@
                     frequency 60s;
                 }
             }
+            field instance-name {
+                sensor bgp-netconf {
+                    where "peer-cfg-rti =~ /{{routing-instance-name}}/";
+                    path peer-cfg-rti;
+                }
+                sensor bgp-oc{
+                    where "/network-instances/network-instance/@name =~ /{{routing-instance-name}}/";
```
gaddevijay, line 32: This might have between releases. Please get the confirmation for Vikram

gaddevijay (changes requested):

juniper_official/routing/check-bgp-neighbor-prefixes.rule
gvarao: Verified and updated with minimum junos and evo support in rules metadata

gaddevijay (approved):

PR #1009, opened 2025-02-07
feat: interface health anomaly fields and term

rahul-kriplani-jnpr:
- in-mbps
- out-mbps
- in-errors-count
- out-errors-count
- fec-corrected
- fec-uncorrected
- input-crc-errors
- framing-errros
- output-crc-errors
- optical-temp
- optics-tx-power
- optics-rx-power

PR #1016, opened 2025-02-24
feat: interface health anomaly fields and term

rahul-kriplani-jnpr:
- in-mbps
- out-mbps
- optical-temp
- optics-tx-power
- optics-rx-power

### DIFF
--- a/juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
+++ b/juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
@@ -213,6 +213,24 @@ healthbot {
                 type integer;
                 description "Interface speed";
             }
+            field in-mbps-anomaly {
+                formula {
+                    jaiml-anomaly-detection {
+                        field-name "$in-mbps";
+                    }
+                }
+                type integer;
+                description "JAIML anomaly detection on input traffic rate";
+            }
+            field out-mbps-anomaly {
+                formula {
+                    jaiml-anomaly-detection {
+                        field-name "$out-mbps";
+                    }
+                }
+                type integer;
+                description "JAIML anomaly detection on output traffic rate";
+            }
             /*
              * Anomaly detection logic.
              */			
@@ -224,9 +242,10 @@ healthbot {
                 /*
                  * Sets color to green when in-util is below low-threshold
                  */				
-                term interface-in-traffic-normal {
+                term is-interface-in-traffic-normal {
                     when {
                         less-than "$in-util" "$low-threshold";
+                        less-than-or-equal-to "$in-mbps-anomaly" 0;
                     }				
                     then {
                         status {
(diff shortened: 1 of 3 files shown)


juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
```diff
@@ -224,17 +242,33 @@ healthbot {
                 /*
                  * Sets color to green when in-util is below low-threshold
                  */				
-                term interface-in-traffic-normal {
+                term is-interface-in-traffic-normal {
```
rahul-kriplani-jnpr, line 245: I have made the term name consistent with others in the trigger

rahul-kriplani-jnpr:
Have added the new terms by taking the reference provided.

I had a doubt, is this term required?
https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L167
aren't we already checking this in
https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L135
is this to handle when jaiml feature is not enabled and we don't get anomaly status?

rahul-kriplani-jnpr:
reference to the closed MR: https://github.com/Juniper/healthbot-rules/pull/1009

juniper_official/interface/check-interface-in-out-errors-traffic-state-flaps.rule
```diff
@@ -224,17 +242,33 @@ healthbot {
...
                     }
-                }				
+                }
+                /*
+                 * Sets color to yellow when in-mbps is not in between predicted range.
+                 */
+                term is-interface-in-traffic-anomalous {
+                    when {
+                        less-than "$in-util" "$low-threshold";
+                        equal-to "$in-mbps-anomaly" 1;
```
rahul-kriplani-jnpr, line 263: the anomaly status will not be greater than one. Should we still add any condition to handle values greater than 1?

gaddevijay: Yes. this is required

rahul-kriplani-jnpr: at this point AIOPs doesn't have anomaly > 1

rahul-kriplani-jnpr:
> Have added the new terms by taking the reference provided.
> 
> I had a doubt, is this term required?
> 
> https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L167
> 
> 
> aren't we already checking this in
> https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L135
> 
> 
> is this to handle when jaiml feature is not enabled and we don't get anomaly status?

I believe this is required for the triggers I have updated, right? @gaddevijay

rahul-kriplani-jnpr:
> > Have added the new terms by taking the reference provided.
> > I had a doubt, is this term required?
> > https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L167
> > 
> > aren't we already checking this in
> > https://github.com/Juniper/healthbot-rules/blob/1548b46c104614909ad5627b007657988305a11a/juniper_official/chassis/check-routing-engine-temperature.rule#L135
> > 
> > is this to handle when jaiml feature is not enabled and we don't get anomaly status?
> 
> I believe this is required for the triggers I have updated, right? @gaddevijay

updated this for the interfaces rules.

gaddevijay (approved):

PR #1022, opened 2025-04-10
Updated rule properties for check-psm-power-usage-state rule

vvikramb:
power usage is supported now on mx204, but temperature is still not supported.

Temperature is supported for mx240,mx480 and mx960.


nmv06:
Jenkins please retry a build

gaddevijay (approved):

PR #1026, opened 2025-05-05
added meta for plots for interface rules

vvikramb:
Added meta for following interface rules :

check-optical-signal-loss-fec-tx-rx-power
check-optical-temp-thresholds
check-interface-fec-crc-framing-errors


juniper_official/interface/check-optical-temp-thresholds.rule
```diff
@@ -61,6 +61,18 @@ healthbot {
                 }
                 type integer;
                 description "Temperature of the Optical Module in degrees Centigrade.";
+                plots temperature-plots {
+                    unit degree-celsius;
```
gaddevijay: Can we just add celsius?

gaddevijay (approved):

PR #1031, opened 2025-05-09
removed thresholds from plots

nchiru:
chassis/check-fan-state-rpm.rule
	chassis/check-psm-power-usage-state.rule
	chassis/check-routing-engine-temperature.rule
	chassis/check-system-power-usage-temp-state.rule
	interface/check-interface-fec-crc-framing-errors.rule
	interface/check-interface-in-out-errors-traffic-state-flaps.rule
	interface/check-optical-signal-loss-fec-tx-rx-power.rule
	interface/check-optical-temp-thresholds.rule
	linecard/check-fpc-cpu-memory-state-temp.rule
	linecard/check-pfe-discards.rule
	system/check-ntp-synchronization-status.rule
	system/check-system-cpu-memory.rule

### DIFF
--- a/juniper_official/chassis/check-fan-state-rpm.rule
+++ b/juniper_official/chassis/check-fan-state-rpm.rule
@@ -81,17 +81,9 @@ healthbot {
                 }
                 type integer;
                 description "RPM percentage";
-                plots rpm-plots {
+                plots rpm {
                     unit percentage;
                     triggers rpm-percent;
-                    thresholds high-threshold {
-                        field-name "$high-threshold";
-                        severity critical;
-                    }
-                    thresholds low-threshold {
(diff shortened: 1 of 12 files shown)


juniper_official/chassis/check-fan-state-rpm.rule
```diff
@@ -81,17 +81,9 @@ healthbot {
                 }
                 type integer;
                 description "RPM percentage";
-                plots rpm-plots {
+                plots rpm {
                     unit percentage;
                     triggers rpm-percent;
-                    thresholds high-threshold {
```
gaddevijay, line 87: Do not remove all the thresholds.
Threshold should be removed for fec-uncorrected plot.

PR #1039, opened 2025-05-21
Added alert type to oam,lldp,ifl,bfd,ospf3 rules

vvikramb:
Added alert type to oam,lldp,ifl,bfd,ospf3 rules

check-bfd-session-state
get-lfm-information
check-lldp-session
check-ospf3-neighbor-state
check-ifl-state

### DIFF
--- a/juniper_official/interface/check-ifl-state.rule
+++ b/juniper_official/interface/check-ifl-state.rule
@@ -154,7 +154,7 @@ healthbot {
              * Anomaly detection logic.
              */
             trigger sub-interface-link-state {
-                alert-type interfaces.state.sub-interface-link-state;			
+                alert-type interfaces.link-failure.sub-interface-link-state;			
                 synopsis "Sub-interface status check kpi";
                 description "Sets health based on ifl state";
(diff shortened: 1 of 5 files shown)


juniper_official/interface/check-ifl-state.rule
```diff
@@ -154,6 +154,7 @@ healthbot {
              * Anomaly detection logic.
              */
             trigger sub-interface-link-state {
+                alert-type interfaces.state.sub-interface-link-state;
```
gaddevijay: interfaces.link_failure.interface-link-state. This is the alert type for IFD can we follow similar hierarchy?

gaddevijay: Also does it make sense to convert _ to - in here?

gaddevijay (changes requested):
please check the comments

juniper_official/interface/check-ifl-state.rule
vvikramb: done

gaddevijay (approved):

PR #1042, opened 2025-06-24
IGP updated ruleprops

vsg26:
- **Default Sensor Addition:**  
  - Added a default sensor for both Junos and JunosEvolved platforms to all single sensor path rules.

- **Multisensor Path Rule Update:**  
  - For the `collect-isis-lsdb-tlv.rule` (a multisensor path rule), added support for several additional platforms along with a default sensor.

### DIFF
--- a/juniper_official/routing/collect-isis-counters.rule
+++ b/juniper_official/routing/collect-isis-counters.rule
@@ -67,32 +67,32 @@
                             products MX {
                                 sensors isis-oc;
                                 platforms MX204 {
-                                    releases 22.2R3 {
+                                    releases 20.2R2 {
                                         release-support min-supported-release;
                                     }
                                 }
                                 platforms MX240 {
-                                    releases 23.2R2 {
+                                    releases 20.2R2 {
                                         release-support min-supported-release;
                                     }
                                 }
                                 platforms MX304 {
-                                    releases 23.2R2 {
+                                    releases 20.2R2 {
                                         release-support min-supported-release;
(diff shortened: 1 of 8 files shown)


gaddevijay (approved):

juniper_official/routing/collect-isis-counters.rule
```diff
@@ -63,73 +63,88 @@
                 supported-devices {
                     juniper {
                         operating-system junos {
+                            sensors isis-oc;
                             products MX {
                                 sensors isis-oc;
                                 platforms MX204 {
-                                    releases 22.2R3 {
+                                    releases 20.2R2 {
```
gaddevijay, line 70: We support release from 22.4. I hope you know that

gaddevijay (changes requested):

juniper_official/routing/collect-isis-counters.rule
vsg26: We want to keep 20.2R2 and we want to receive the data if there is 20.2R2 device if it happens to be there in the network.
 
If we have not qualified that release, thats fine, we dont want to drop the data just because it was not qualified. If the path was different and anomaly detection doesnt work for 20.2R2 it's fine but we want to be liberal in accepting data.

vsg26: also we are aware of not having support for rel <22.4

gaddevijay (approved):

PR #1043, opened 2025-07-21
feat: default trigger to store ts-input-packets-pps ts-output-packets-pps to tsdb

rahul-kriplani-jnpr:
Added a default trigger to store ts-input-packets-pps, ts-output-packets-pps and ts-drop-packets-pps in TSDB

### DIFF
--- a/juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
+++ b/juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
@@ -32,6 +32,15 @@ healthbot {
                 type unsigned-integer;
                 description "PFE output stats";
             }
+            field ts-drop-packets-pps {
+                formula {
+                    eval {
+                        expression "max(($ts-input-packets-pps - $ts-output-packets-pps), 0)";
+                    }
+                }
+                type unsigned-integer;
+                description "PFE drop stats";				
+            }
             field fpc {
                 sensor pfestats {
                     path "/components/component/@name";
--- a/juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
+++ b/juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
@@ -32,15 +32,6 @@ healthbot {
                 type unsigned-integer;
                 description "PFE output stats";
             }
-            field ts-drop-packets-pps {
-                formula {
-                    eval {
-                        expression "($ts-input-packets-pps - $ts-output-packets-pps) * (($ts-input-packets-pps - $ts-output-packets-pps) > 0)";
-                    }
-                }
-                type unsigned-integer;
-                description "PFE drop stats";				
-            }
             field fpc {
                 sensor pfestats {
                     path "/components/component/@name";
@@ -56,7 +47,7 @@ healthbot {
                     then {
                         status {
                             color green;
-                            message "$ts-input-packets-pps $ts-output-packets-pps $ts-drop-packets-pps";
+                            message "$ts-input-packets-pps $ts-output-packets-pps";
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


rahul-kriplani-jnpr:
@vvikramb, @gaddevijay, @sanjeevm-jsi Please review.

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
```diff
@@ -32,13 +32,35 @@ healthbot {
                 type unsigned-integer;
                 description "PFE output stats";
             }
+            field ts-drop-packets-pps {
+                formula {
+                    eval {
+                        expression "($ts-input-packets-pps - $ts-output-packets-pps) * (($ts-input-packets-pps - $ts-output-packets-pps) > 0)";
```
gaddevijay: How does this give drop pps? I don't think this is an accurate calculation

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
```diff
@@ -32,13 +32,35 @@ healthbot {
...
+                    }
+                }
+                type unsigned-integer;
```
gaddevijay: We don't store unsigned integer

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
```diff
@@ -32,13 +32,35 @@ healthbot {
...
+                description "PFE drop stats";				
+            }
             field fpc {
                 sensor pfestats {
                     path "/components/component/@name";
                 }
                 type string;
                 description "FPC name";
             }
+            trigger pfe-input-output-stats {
```
gaddevijay, line 42: Why does this look like a dummy trigger to me? Is AIML consuming any trigger data?

sanjeevm-jsi (approved):

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
gaddevijay: Why this needs to be stored? What is the purpose of this?

gaddevijay (changes requested):

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
rahul-kriplani-jnpr: Yes, its a dummy commit as  we wanted to store ts-input-packets-pps and ts-output-packets-pps to TSDB to plot the Traffic Loss graph (in Routing and MPLS accordion) in real time like we plot the graphs for Hardware and Interfaces accordion.

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
rahul-kriplani-jnpr: I wanted the calculate max((ts-input-packets-pps - ts-output-packets-pps), 0) to achieve this I added such expression.
Also this is an outdated change, not part of the PR now as we are calculating this elsewhere.

juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule
rahul-kriplani-jnpr: I saw this being used for other fields (https://github.com/Juniper/healthbot-rules/blob/b79538770047d438528c3fb64e850bccfd684683/juniper_official/Solutions/Traffic-Blackhole/pfe-stats.rule#L24) and as this was being derived from those fields I used the same type.
Also this is an outdated change, not part of the PR now.

PR #1044, opened 2025-08-13
Update collect-logical-interface-stats.rule to support Cisco

madhuranj:
The requirement is to collect the logical interface traffic statistics from Cisco devices, for which the existing rule "collect-logical-interface-stats.rule" has been updated with Cisco specific openconfig sensor and platform/models

Note:
* Testing is done on virtual machine having the IOS-XR 7.5.1, 7.8.1 and 24.3.1
* The rule has both "XRV Appliance" and "XRv" because EMS has issues supporting "XRv"

### DIFF
--- a/juniper_official/interface/collect-logical-interface-stats.rule
+++ b/juniper_official/interface/collect-logical-interface-stats.rule
@@ -2,6 +2,7 @@
  * Collect Logical Interface traffic statistics for pathfinder dynamic topology.
  *
  * Req 60: https://paragon-automation.atlassian.net/browse/REQ-60
+ * Req 1351: https://paragon-automation.atlassian.net/browse/REQ-1351
  *
  * Two inputs control detection
  *
@@ -28,6 +29,13 @@
                     frequency 60s;
                 }
             }
+            sensor cisco-oc-ifl {
+                description "Cisco open-config Logical interface sensor definition";
+                open-config {
+                    sensor-name /interfaces/interface/subinterfaces/subinterface;
+                    frequency 60s;
+                }
+            }
             field out-bps {
                 formula {
                     rate-of-change {
@@ -41,11 +49,17 @@
                 sensor interfaces-oc {
                     path /interfaces/interface/subinterfaces/subinterface/state/counters/out-octets;
                 }
+                sensor cisco-oc-ifl {
+                    path /interfaces/interface/subinterfaces/subinterface/state/counters/out-octets;
+                }
                 type integer;
                 description "Subinterface egress octets";
             }
             field out-pkts {
                 sensor interfaces-oc {
+                    path /interfaces/interface/subinterfaces/subinterface/state/counters/out-pkts;
+                }
+                sensor cisco-oc-ifl {
                     path /interfaces/interface/subinterfaces/subinterface/state/counters/out-pkts;
                 }
                 type integer;
@@ -72,11 +86,17 @@
                 sensor interfaces-oc {
                     path /interfaces/interface/subinterfaces/subinterface/state/counters/in-octets;
                 }
+                sensor cisco-oc-ifl {
+                    path /interfaces/interface/subinterfaces/subinterface/state/counters/in-octets;
+                }
                 type integer;
                 description "Subinterface ingress octets";
             }
             field in-pkts {
                 sensor interfaces-oc {
+                    path /interfaces/interface/subinterfaces/subinterface/state/counters/in-pkts;
+                }
+                sensor cisco-oc-ifl {
                     path /interfaces/interface/subinterfaces/subinterface/state/counters/in-pkts;
                 }
                 type integer;
@@ -95,6 +115,10 @@
                     where "/interfaces/interface/@name =~ /{{input-ifd-name}}/";
                     path "/interfaces/interface/@name";
                 }
+                sensor cisco-oc-ifl {
+                    where "/interfaces/interface/@name =~ /{{input-ifd-name}}/";
+                    path "/interfaces/interface/@name";
+                }
                 type string;
                 description "Interfaces to be monitored";
             }
@@ -103,8 +127,28 @@
                     where "/interfaces/interface/subinterfaces/subinterface/@index =~ /{{input-ifl-no}}/";
                     path "/interfaces/interface/subinterfaces/subinterface/@index";
                 }
+                sensor cisco-oc-ifl {
+                    where "/interfaces/interface/subinterfaces/subinterface/@index =~ /{{input-ifl-no}}/";
+                    path "/interfaces/interface/subinterfaces/subinterface/@index";
+                }
                 type string;
                 description "Interfaces to be monitored";
+            }
+            field vendor {
+                sensor interfaces-oc {
+                    path juniper_logical_interface_vendor_dummy_path;
+                    data-if-missing {
+                        value "juniper";
+                    }
+                }
+                sensor cisco-oc-ifl {
+                    path cisco_logical_interface_vendor_dummy_path;
+                    data-if-missing {
+                        value "cisco";
+                    }              
+                }
+                type string;                
+                description "Indicates telemetry vendor";
             }
             variable input-ifd-name {
                 value .*;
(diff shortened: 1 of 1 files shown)


vvikramb:
Please check : 

Can we use the same sensor for cisco devices if sensor paths are the same? 

Below points to consider if seperate sensor for cisco:

1.Sensor path might be same for other Vendors like Nokia so it is better to use generic name for sensor name instead of "cisco-oc-ifl".
2.For field "Vendor" , data-if-missing should not be "cisco" if same  sensor is being used for other vendors also.
3."input-ifd-name" and "input-ifl-no" can be  "ifd-name" and "ifl-no".
4.Do we need bps and pps both?
5.In rule-properties ,Sensor paths are defined under platforms and releases, it can be removed from under releases if there is no change.
6.We have qualified NCS57C3 and NCS5508 for insights, but NCS5504 and NCS5508 is mentioned, are the models different for PF? 

Better to have different rule for non-junos as below config is not applicable for others : 

    version 1;
    contributor juniper;
    category basic;
    is-scaling-rule {
        description "Fields: interface-name ; Directly impacted by number of interfaces and IFLs in each network device";
    }
    supported-healthbot-version 4.2.0;

Please check all rules are getting committed using paragon cli.

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -240,4 +361,4 @@
             }
```
raghurampai, line 294: Please check if closing braces are matching

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -103,9 +127,29 @@
...
+                sensor cisco-oc-ifl {
+                    path cisco_logical_interface_vendor_dummy_path;
+                    data-if-missing {
+                        value "cisco";
+                    }              
+                }
+                type string;                
+                description "Indicates telemetry vendor";
+            }
             variable input-ifd-name {
```
raghurampai: can this be just “ifd-name” to avoid confusion with IO?

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -28,6 +29,13 @@
                     frequency 60s;
                 }
             }
+            sensor cisco-oc-ifl {
+                description "Cisco open-config Logical interface sensor definition";
+                open-config {
+                    sensor-name /interfaces/interface/subinterfaces/subinterface;
```
raghurampai: For both juniper and cisco, field paths look identical. If so, will same sensor path(with “/state”) and rule work for cisco? this is just to avoid confusion and support precedence in rule if needed say for other vendors in future.

madhuranj:
Hi VijayVikram, 
Thank you for going through the rule and providing your inputs, please find my response below starting with [MMN].

> Can we use the same sensor for cisco devices if sensor paths are the same?

[MMN]: Even though the sensor path is the same, need a way to identify the vendor for which the data is being received, hence need individual sensor for each vendor

>1.Sensor path might be same for other Vendors like Nokia so it is better to use generic name for sensor name instead of "cisco-oc-ifl".

[MMN]: The intention is to have vendor specific sensor name to distinguish the data being received even if the sensor path is same because based on vendor specifc path the value for the "vendor" field is populated

> 2. For field "Vendor" , data-if-missing should not be "cisco" if same sensor is being used for other vendors also.

[MMN]: The vendor information is needed by the backend code to check the feature flag, since the same rule is used for both Juniper and Cisco, the backend code needs the vendor name as a input data point hence the rule is using the vendor specific sensor with dummy path to populate the vendor name

> 3."input-ifd-name" and "input-ifl-no" can be "ifd-name" and "ifl-no".

[MMN]: These are the existing names and not newly added names, any change to these will require changes to the backend code that creates devices and involves significate amount of effort to change and test hence and cannot be addressed in the current release, will be addressed when a new vendor is added

> 4.Do we need bps and pps both?

[MMN]: Yes PathFinder uses need both bps and pps

> 5.In rule-properties ,Sensor paths are defined under platforms and releases, it can be removed from under releases if there is no change.

[MMN]: Ok will be addressed if any change is done to the file in this MR, else will be done in the next release

> 6.We have qualified NCS57C3 and NCS5508 for insights, but NCS5504 and NCS5508 is mentioned, are the models different for PF?

[MMN]: These platforms to be supported were received as input from PLM

> 7. Better to have different rule for non-junos as below config is not applicable for others:

[MMN]: This section was added previously and is not part of the current changes, the Cisco sensors were added to the existing rule based on the information received during the discussion with Insights team before the development started for Cisco telemetry support. It will be a very big effort to create different rules for Juniper and Cisco at this point in the release cycle that may result in PathFinder not being able to support Cisco telemetry in 2.6 release

> 8. Please check all rules are getting committed using paragon cli.

[MMN]: The json rules used create the rules in the setup shared over email were generated using the rules that where committed through the paragon cli editor.

raghurampai:
> Hi VijayVikram, Thank you for going through the rule and providing your inputs, please find my response below starting with [MMN].
> 
> > Can we use the same sensor for cisco devices if sensor paths are the same?
> 
> [MMN]: Even though the sensor path is the same, need a way to identify the vendor for which the data is being received, hence need individual sensor for each vendor
> 
> > 1.Sensor path might be same for other Vendors like Nokia so it is better to use generic name for sensor name instead of "cisco-oc-ifl".
> 
> [MMN]: The intention is to have vendor specific sensor name to distinguish the data being received even if the sensor path is same because based on vendor specifc path the value for the "vendor" field is populated
> 
> > 2. For field "Vendor" , data-if-missing should not be "cisco" if same sensor is being used for other vendors also.
> 
> [MMN]: The vendor information is needed by the backend code to check the feature flag, since the same rule is used for both Juniper and Cisco, the backend code needs the vendor name as a input data point hence the rule is using the vendor specific sensor with dummy path to populate the vendor name
> 
> > 3."input-ifd-name" and "input-ifl-no" can be "ifd-name" and "ifl-no".
> 
> [MMN]: These are the existing names and not newly added names, any change to these will require changes to the backend code that creates devices and involves significate amount of effort to change and test hence and cannot be addressed in the current release, will be addressed when a new vendor is added
> 
> > 4.Do we need bps and pps both?
> 
> [MMN]: Yes PathFinder uses need both bps and pps
> 
> > 5.In rule-properties ,Sensor paths are defined under platforms and releases, it can be removed from under releases if there is no change.
> 
> [MMN]: Ok will be addressed if any change is done to the file in this MR, else will be done in the next release
> 
> > 6.We have qualified NCS57C3 and NCS5508 for insights, but NCS5504 and NCS5508 is mentioned, are the models different for PF?
> 
> [MMN]: These platforms to be supported were received as input from PLM
> 
> > 7. Better to have different rule for non-junos as below config is not applicable for others:
> 
> [MMN]: This section was added previously and is not part of the current changes, the Cisco sensors were added to the existing rule based on the information received during the discussion with Insights team before the development started for Cisco telemetry support. It will be a very big effort to create different rules for Juniper and Cisco at this point in the release cycle that may result in PathFinder not being able to support Cisco telemetry in 2.6 release
> 
> > 8. Please check all rules are getting committed using paragon cli.
> 
> [MMN]: The json rules used create the rules in the setup shared over email were generated using the rules that where committed through the paragon cli editor.

Hi Madhura,
On point 1 and 2 above- the mechanism of supporting different sensor for every vendor may be difficult to manage(for every vendor you will need to update the rule even if openconfig model is supported). If the intention is to support a field which needs to have a fixed vendor value, then you can consider using variables and constant (example- https://ssd-git.juniper.net/mistsys/custom-artifacts/-/blob/main/juniper-engineering/custom-kpis/interfaces/check-interface-in-out-errors-traffic-state-flaps.json?ref_type=heads#L99). When deploying the rule, this value can be passed as deployment type variable.
Thanks!

raghurampai (approved):
Looks good overall, apart from the suggestion provided.

juniper_official/interface/collect-logical-interface-stats.rule
gaddevijay: having /state makes sense here based on the what we are monitoring for Cisco.

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -103,9 +127,29 @@
...
                     path "/interfaces/interface/subinterfaces/subinterface/@index";
                 }
+                sensor cisco-oc-ifl {
+                    where "/interfaces/interface/subinterfaces/subinterface/@index =~ /{{input-ifl-no}}/";
+                    path "/interfaces/interface/subinterfaces/subinterface/@index";
+                }
                 type string;
                 description "Interfaces to be monitored";
             }
+            field vendor {
```
gaddevijay: Not telemetry processing. Should be moved to PF pipelines

gaddevijay (changes requested):

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: The braces are matching, also rule was used to generate the json file used during testing and it can be found in the setup shared over email

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: Have renamed it to "inp-interface-name" to keep it same as the field name but with the prefix "inp-" indicating that it is user input.

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: Cisco specific sensor is removed

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: Cisco specific sensor is removed

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -92,30 +96,42 @@
...
+            }             
+            field device-vendor {
+                constant {
+                    value "{{inp-device-vendor}}";
+                }
+                type string;                
+                description "Indicates device vendor";
             }
-            variable input-ifd-name {
+            variable inp-interface-name {
```
gaddevijay: Changing of variable names will cause upgrade failures if the rules is already instantiated on older release

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -92,30 +96,42 @@
...
                 value .*;
                 description "Interfaces to be monitored, regular expression, eg 'ge-.*'";
                 type string;
             }
-            variable input-ifl-no {
+            variable inp-sub-interface-index {
```
gaddevijay: Changing of variable names will cause upgrade failures if the rules is already instantiated on older release

gaddevijay (changes requested):

gaddevijay (approved):

PR #1045, opened 2025-08-13
Update collect-physical-interface-stats.rule to support Cisco

madhuranj:
The requirement is to collect the physical interface traffic statistics from Cisco devices, for which the existing rule "collect-physical-interface-stats.rule" has been updated with Cisco specific openconfig sensor and platform/models

Note:

    Testing is done on virtual machine having the IOS-XR 7.5.1, 7.8.1 and 24.3.1
    The rule has both "XRV Appliance" and "XRv" because EMS has issues supporting "XRv"

### DIFF
--- a/juniper_official/interface/collect-physical-interface-stats.rule
+++ b/juniper_official/interface/collect-physical-interface-stats.rule
@@ -4,30 +4,26 @@
  * Req 60: https://paragon-automation.atlassian.net/browse/REQ-60
  * Req 1351: https://paragon-automation.atlassian.net/browse/REQ-1351
  *
- * One input control detection
+ * Two input control detection
  *
- *   1) input-ifd-name, is a regular expression that matches the interfaces
+ *   1) inp-interface-name, is a regular expression that matches the interfaces
  *      that you would like to monitor.  By default it's '.*', which matches
  *      all interfaces. Use something like 'ge.*' to match only gigabit
- *      ethernet interfaces. 
+ *      ethernet interfaces.
+ *
+ *   2) inp-device-vendor, indicates the device vendor is empty by default and needs to be filled during
+ *      rule deployment based on device. 
  */
  healthbot {
     topic interfaces {
         rule collect-physical-interface-stats {
-            keys [ interface-name ];
+            keys interface-name;
             description "Collects physical interface traffic stats periodically";
             sensor interfaces-oc {
                 synopsis "Interface open-config sensor definition";
                 description "Interface open-config sensor definition";
                 open-config {
                     sensor-name /interfaces/interface/state;
-                    frequency 60s;
-                }
-            }
-            sensor cisco-oc-ifd {
-                description "Cisco open-config Physical interface sensor definition";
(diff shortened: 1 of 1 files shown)


raghurampai (approved):
Looks good. Can check and consider similar suggestion for optimization as ifl rule.

juniper_official/interface/collect-physical-interface-stats.rule
```diff
@@ -90,9 +110,29 @@
...
                     path "/interfaces/interface/@name";
                 }
+                sensor cisco-oc-ifd {
+                    where "/interfaces/interface/@name =~ /{{input-ifd-name}}/";
+                    path "/interfaces/interface/@name";
+                }
                 type string;
                 description "Interfaces to be monitored";
             }
+            field vendor {
```
gaddevijay: Should be removed and needs to be handled in PF pipeline

juniper_official/interface/collect-physical-interface-stats.rule
```diff
@@ -23,6 +24,13 @@
                     frequency 60s;
                 }
             }
+            sensor cisco-oc-ifd {
+                description "Cisco open-config Physical interface sensor definition";
+                open-config {
+                    sensor-name /interfaces/interface/state/counters;
```
gaddevijay: I think it's better to subscribe to /interfaces/interface/state/. Please sync up with Vikram as there will be observability rules on this, we would like to align from product perspective to minimise data collection.

gaddevijay (changes requested):

juniper_official/interface/collect-physical-interface-stats.rule
madhuranj: Dummy path replaced with constants

juniper_official/interface/collect-physical-interface-stats.rule
madhuranj: Cisco specific sensor has been removed

juniper_official/interface/collect-physical-interface-stats.rule
gaddevijay: This needs to be removed from Rule and moved to PF pipeline.

juniper_official/interface/collect-physical-interface-stats.rule
```diff
@@ -87,17 +91,29 @@
...
             field interface-name {
                 sensor interfaces-oc {
-                    where "/interfaces/interface/@name =~ /{{input-ifd-name}}/";
+                    where "/interfaces/interface/@name =~ /{{inp-interface-name}}/";
                     path "/interfaces/interface/@name";
                 }
                 type string;
                 description "Interfaces to be monitored";
-            }
-            variable input-ifd-name {
```
gaddevijay: Changing of variable names will cause upgrade failures if the rules is already instantiated on older release

gaddevijay (changes requested):

gaddevijay (approved):

PR #1047, opened 2025-08-13
 Update collect-sr-color-lsp-stats.rule to support Cisco

madhuranj:
The requirement is to collect the SR Color LSP traffic statistics from Cisco devices, for which the existing rule "collect-sr-color-lsp-stats.rule" has been updated with Cisco specific openconfig sensor and platform/models.
The name of the Juniper SR LSP is computed using two fields, where as the name for Cisco is received in a single field hence a new field has been added to the key which contains the Cisco SR LSP name.

Note:

- Testing is done on virtual machine having the IOS-XR 7.5.1, 7.8.1 and 24.3.1
- The rule has both "XRV Appliance" and "XRv" because EMS has issues supporting "XRv"

### DIFF
--- a/juniper_official/routing/collect-sr-color-lsp-stats.rule
+++ b/juniper_official/routing/collect-sr-color-lsp-stats.rule
@@ -2,24 +2,26 @@
  * Collect SR Color LSP traffic statistics for pathfinder dynamic topology.
  *
  * Req 1936: https://paragon-automation.atlassian.net/browse/REQ-1936
- * Req 1351: https://paragon-automation.atlassian.net/browse/REQ-1351
  *
- * Two input control detection
+ * Three input control detection
  *
- *   1) input-to-address, is a regular expression that matches the SR color lsp 
+ *   1) inp-lsp-destination, is a regular expression that matches the SR color lsp 
  *      destination address that must be monitored.  By default it's '.*', which matches
  *      all destination address. Use something like '11.0.0*' to match only destination 
  *      address starting with '11.0.0'.  
  *
- *   2) input-color, is a regular expression that matches the color of SR color lsp 
+ *   2) inp-lsp-color, is a regular expression that matches the color of SR color lsp 
  *      that must be monitored.  By default it's '.*', which matches
  *      all colors. Use something like '10*' to match LSPs starting with color '10'.
+ *
+ *   3) inp-device-vendor, indicates the device vendor is empty by default and needs to be filled during
+ *      rule deployment based on device.  
  *
  */
 healthbot {
     topic routing.spring {
         rule collect-sr-color-lsp-stats {
-            keys [ lsp-name lsp-cfg-name ];
+            keys [ lsp-name ];
             description "Collects SR color LSP stats periodically";
             sensor jnpr-sr-color-lsp {
                 open-config {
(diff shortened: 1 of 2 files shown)


raghurampai (approved):

juniper_official/routing/collect-sr-color-lsp-stats.rule
```diff
@@ -63,7 +89,37 @@ healthbot {
...
+                    }
+                }
+                sensor cisco-oc-sr-color-lsp {
+                    where "/interfaces/interface/@name =~ /{{input-cisco-sr-color-lsp-name}}/";
+                    path "/interfaces/interface/@name";
+                }                
+                type string;
+                description "Interface name";
+            }
+            field vendor {
```
gaddevijay: Please use the vendor from the inventory and not overload the telemetry processing

juniper_official/routing/collect-sr-color-lsp-stats.rule
```diff
@@ -18,41 +19,66 @@
...
             }
             field lsp-color {
                 sensor jnpr-sr-color-lsp {
                     where "/mpls/signaling-protocols/segment-routing/sr-te-ip-policies/sr-te-ip-policy/@color =~ /{{input-color}}/";
                     path "/mpls/signaling-protocols/segment-routing/sr-te-ip-policies/sr-te-ip-policy/@color";
-                }                
+                }
+                sensor cisco-oc-sr-color-lsp {
+                    path cisco_lsp_color_dummy_path;
+                    data-if-missing {
```
gaddevijay: data if missing makes the data to buffer for 1.5 times the sensor frequency. Which means though we know that the data would come of this point we will have to buffer it for 90s which is very inefficient way to process at scale.

Ideal case would be is to write the 2 rules and for different vendors where key attributes like this differ.

juniper_official/routing/collect-sr-color-lsp-stats.rule
```diff
@@ -129,13 +190,13 @@ healthbot {
...
                                 platforms PTX1000 {
-                                    releases 22.1R1 {
+                                    releases 22.4R1 {
                                         sensors jnpr-sr-color-lsp;
                                         release-support min-supported-release;
                                     }
                                 }
                                 platforms PTX10000  {
-                                    releases 22.1R1 {
+                                    releases 22.4R1 {
```
gaddevijay: Does this mean that 22.1R1 does not support this sensor/rule. If so PF is not instantiating it?

gaddevijay (changes requested):

juniper_official/routing/collect-sr-color-lsp-stats.rule
madhuranj: Using the "vendor" data from inventory can lead to errors because, it gives inconsistent response, that is, it does not give vendor info for Juniper device, if this info is used, then the code will be forced to assume that missing vendor implies the device vendor is Juniper, which may cause errors if there is bug and the inventory API is not populating the vendor info for non Juniper devices.
This field has been updated to use constants instead of dummy path

juniper_official/routing/collect-sr-color-lsp-stats.rule
madhuranj: New rule "collect-srte-interface-stats.rule" add to handle the Cisco SR Color LSP traffic that is available as SRTE interface traffic.

juniper_official/routing/collect-sr-color-lsp-stats.rule
madhuranj: Received confirmation from Bala that Junos fully supports GNMI Dialout only from 22.4 hence updated the same.
PF does not have the requirement to restrict the rule deployment based on the model version.

gaddevijay: As of today for backward compatibility reasons for single sensor rules, Insights do not restrict applying the rules based on versions. This is something we are working on to exclude non supported versions.

If the rule is instantiated and the sensor is not supported by device, then device will keep terminating the connections and telemetry collection will get impacted. It's better to have this check until above features is present in Insights.

Since 22.4R1 is from when dial support is officially available, it should be safe to assume that all devices versions > 22.4R1 for now.

juniper_official/routing/collect-sr-color-lsp-stats.rule
```diff
@@ -83,16 +86,28 @@ healthbot {
...
+            }            
+            field device-vendor {
+                constant {
+                    value "{{inp-device-vendor}}";
+                }
+                type string;                
+                description "Indicates device vendor";
             }
-            variable input-to-address {
+            variable inp-lsp-destination {
```
gaddevijay: variable name changes will impact upgrade scenarios

juniper_official/routing/collect-sr-color-lsp-stats.rule
```diff
@@ -83,16 +86,28 @@ healthbot {
...
                 value .*;
                 description "SR color LSP destination address to monitor, regular expression, eg '11.0.0.*|10.52.135.*'";
                 type string;
             }
-            variable input-color {
+            variable inp-lsp-color {
```
gaddevijay: Changing of variable names will cause upgrade failures if the rules is already instantiated on older release

gaddevijay (changes requested):

gaddevijay (approved):

PR #1048, opened 2025-08-13
Update REQ and supported versions in collect-sr-lsp-stats.rule

madhuranj:
The rule "collect-sr-lsp-stats.rule" was missing the REQ number, the same has been added, also the minimum supported version has been changed to the version where junos supports openconfig fully.

### DIFF
--- a/juniper_official/routing/collect-sr-lsp-stats.rule
+++ b/juniper_official/routing/collect-sr-lsp-stats.rule
@@ -1,7 +1,7 @@
 /*
  * Collect SR LSP traffic statistics for pathfinder dynamic topology.
  *
- * Req <TBD>: TBD 
+ * Req 1936: https://paragon-automation.atlassian.net/browse/REQ-1936 
  *
  * One input control detection
  *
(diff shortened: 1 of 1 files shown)


raghurampai (approved):

juniper_official/routing/collect-sr-lsp-stats.rule
```diff
@@ -76,25 +76,25 @@ healthbot {
                             products MX {
                                 sensors jnpr-sr-lsp;
                                 platforms MX960 {
-                                    releases 22.1R1 {
```
gaddevijay, line 79: Does this change mean that 22.1R1 is not supported on these platforms?
If yes, does PF takes care of not instantiating the rules?

gaddevijay (changes requested):

juniper_official/routing/collect-sr-lsp-stats.rule
madhuranj: Received confirmation from Bala that Junos fully supports GNMI Dialout only from 22.4 hence updated the same.
PF does not have the requirement to restrict the rule deployment based on the model version.

juniper_official/routing/collect-sr-lsp-stats.rule
```diff
@@ -62,11 +65,23 @@ healthbot {
...
+            }           
+            field device-vendor {
+                constant {
+                    value "{{inp-device-vendor}}";
+                }
+                type string;                
+                description "Indicates device vendor";
             }
-            variable input-lsp-name {
+            variable inp-lsp-name {
```
gaddevijay: Changing of variable names will cause upgrade failures if the rules is already instantiated on older release

juniper_official/routing/collect-sr-lsp-stats.rule
```diff
@@ -62,11 +65,23 @@ healthbot {
                 }
                 type integer;
                 description "LSP packet rate";
+            }           
+            field device-vendor {
```
gaddevijay: Needs to move to PF pipeline

gaddevijay (changes requested):

gaddevijay (approved):

PR #1063, opened 2025-10-07
fix: PPF-2110: Add vmx and mx304 special handling for SR LSP

madhuranj:
The change included in this MR are,
1. Update Juniper specific rules to support the Juniper VMX device properties 
2. In the Juniper MX304 device the SR LSP sensor has a bug , that is, the field that should contain the tunnel name is sent in the field named "tunnelName" instead of "tunnel-name", this MR contains the changes to handle this case.

Note: The testing is done only for VMX changes, the testing for MX304 special handling cannot be done by development team as we do not have access to MX304, this SR LSP changes will be tested by QA team.

refs: PPF-2110, PPF-2108

### DIFF
--- a/juniper_official/interface/collect-logical-interface-stats.rule
+++ b/juniper_official/interface/collect-logical-interface-stats.rule
@@ -261,6 +261,15 @@
                                     }
                                 }
                             }
+                            products VMX {
+                                sensors interfaces-oc;
+                                platforms VMX {                                    
+                                    sensors interfaces-oc;
+                                    releases 22.4R1 {
+                                        release-support min-supported-release;
+                                    }
+                                }
+                            }
                         }
                     }                    
                     other-vendor cisco {
--- a/juniper_official/routing/collect-sr-lsp-stats.rule
+++ b/juniper_official/routing/collect-sr-lsp-stats.rule
@@ -22,11 +22,21 @@ healthbot {
                     frequency 60s;
                 }
             }
+            sensor jnpr-sr-lsp-tunnelName {
+                open-config {
+                    sensor-name /junos/services/segment-routing/traffic-engineering/tunnel/ingress/usage/;
+                    frequency 60s;
+                }
+            }
             field lsp-name {
                 sensor jnpr-sr-lsp {
                     where "/mpls/signaling-protocols/segment-routing/sr-te-ingress-tunnel-policies/sr-te-ingress-tunnel-policy/@tunnel-name =~ /{{input-lsp-name}}/";
                     path "/mpls/signaling-protocols/segment-routing/sr-te-ingress-tunnel-policies/sr-te-ingress-tunnel-policy/@tunnel-name";
-                }              
+                }   
+                sensor jnpr-sr-lsp-tunnelName {
+                    where "/mpls/signaling-protocols/segment-routing/sr-te-ingress-tunnel-policies/sr-te-ingress-tunnel-policy/@tunnelName =~ /{{input-lsp-name}}/";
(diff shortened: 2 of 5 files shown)


juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -261,6 +261,15 @@
                                     }
                                 }
                             }
+                            products VMX {
```
raghurampai, line 264: Please confirm the products and platforms deployment name for “VMX” platform, to ensure it matches with the deployment where this is intended to be used. As we know, in internally deployed test vmx’s the platform name is something like MX960 (that is configuration in vmm deployer file) and giving it as “VMX” in rule will not deploy this rule for that platform. 

Same suggestion in other files below.

juniper_official/interface/collect-logical-interface-stats.rule
```diff
@@ -261,6 +261,15 @@
...
+                                sensors interfaces-oc;
+                                platforms VMX {                                    
+                                    sensors interfaces-oc;
+                                    releases 22.4R1 {
+                                        release-support min-supported-release;
```
raghurampai, line 269: Just to note, the “release” version mentioned doesn’t play a role in sensor deployment here(because there is no differentiated sensor configured under this hierarchy). "sensors interfaces-oc” will be deployed for any release version of this platform.

juniper_official/routing/collect-sr-lsp-stats.rule
```diff
@@ -22,25 +22,41 @@ healthbot {
                     frequency 60s;
                 }
             }
+            sensor jnpr-sr-lsp-tunnelName {
```
raghurampai, line 25: Minor suggestion on naming- can avoid mixing “camelCase” to existing dash naming

juniper_official/routing/collect-sr-lsp-stats.rule
```diff
@@ -72,9 +88,16 @@ healthbot {
                 supported-devices {
                     juniper {
                         operating-system junos {
-                            sensors jnpr-sr-lsp;	
+                            sensors jnpr-sr-lsp;
                             products MX {
                                 sensors jnpr-sr-lsp;
+                                platforms MX304 {
+                                    sensors jnpr-sr-lsp-tunnelName;
```
raghurampai, line 95: In this case any version of MX304 will be deployed with "sensors jnpr-sr-lsp-tunnelName”, thus making the sensors defined under the hierarchy "releases 23.4R2” redundant. Please recheck if this is the intended behaviour.

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: This product and platform has been tested with the new network profile added for vJunos, it is a requirement from PLM since vMX is EOL, will share the setup details where this has been tested in email.

juniper_official/interface/collect-logical-interface-stats.rule
madhuranj: Ack

juniper_official/routing/collect-sr-lsp-stats.rule
madhuranj: For now we know only about the tunnelName in 23.4R2, it may or maynot be same in subsequent releases, we are waiting for input from Junos team, based on their information, we may have to add other releases, but for now the intention is to have the sensor "jnpr-sr-lsp-tunnelName" to be used for MX304 for all releases

gaddevijay (approved):

PR #1091, opened 2026-02-16
Updated interface variable value for the interface rules

vvikramb:
Updated interface variable value for the interface rules :

check-ifl-state
check-optical-temp-thresholds
check-optical-signal-loss-fec-tx-rx-power
check-interface-in-out-errors-traffic-state-flaps
check-interface-fec-crc-framing-errors

### DIFF
--- a/juniper_official/interface/check-ifl-state.rule
+++ b/juniper_official/interface/check-ifl-state.rule
@@ -334,6 +334,7 @@ healthbot {
                 value ".*";
                 description "Interfaces to be monitored, regular expression, eg 'ge-.*'";
                 type string;
+                value-type interface;				
             }
             variable ifl-no {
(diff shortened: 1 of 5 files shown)


juniper_official/interface/check-ifl-state.rule
```diff
@@ -360,6 +360,11 @@ healthbot {
                 description "Subinterface input bandwidth";
                 type int;
             }
+            variable interface {
```
gaddevijay: This is not what we should do. there is a value type that needs to be added to IFD variables

gaddevijay (changes requested):

vvikramb:
Jenkins please retry a build

gaddevijay (approved):

PR #1093, opened 2026-02-24
feat: add local-link-id and local-ipv4-address fields to collect-isis…

mohinisaha612:
Added two new fields to collect-isis-lsdb-neighbsysid-v2.rule:

- local-link-id
- local-ipv4-address

### DIFF
--- a/juniper_official/routing/collect-isis-lsdb-neighbsysid-v2.rule
+++ b/juniper_official/routing/collect-isis-lsdb-neighbsysid-v2.rule
@@ -52,6 +52,20 @@
                 type integer;
                 description "state metric value from the lsdb path";
             }
+            field local-link-id {
+                sensor isis-oc {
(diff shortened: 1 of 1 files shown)


vvikramb:
Rule looks good, have you tested on the platforms mentioned in the rule properties, better to test on atleast one junos and junos-evolved model, we have seen some discrepancies earlier across models and also across releases in some cases.

Also you can mention the ticket or Req# if the additional fields added now are not part of Req 1349.

juniper_official/routing/collect-isis-lsdb-neighbsysid-v2.rule
```diff
@@ -52,6 +52,20 @@
                 type integer;
                 description "state metric value from the lsdb path";
             }
+            field local-link-id {
```
gaddevijay, line 55: Please update the REQ number at the top of the file. There is already one REQ. Please append to that

gaddevijay (changes requested):

gaddevijay (approved):

